PR #17, opened 2025-09-29
feat: credential structure and parsing

paolodamico:
- Introduces the base `Credential` structure and claims.
- Introduces basic field element primitives with serialization/deserialization.


crates/world-id-core/src/primitives.rs
```diff
@@ -0,0 +1,150 @@
...
+            if let Some(rest) = s.strip_prefix("0x").or_else(|| s.strip_prefix("0X")) {
+                s = rest.to_string();
+            }
+
+            hex::decode(&s).map_err(D::Error::custom)?
+        } else {
+            <Vec<u8>>::deserialize(deserializer)?
+        };
+        // TODO: Check the expected length?
+        Ok(Self(Fq::from_be_bytes_mod_order(&bytes)))
```
decentralgabe, line 55: this seems important, without it we'll risk modular reduction (e.g. `0x01`, `0x0001`, `0x(p+1)` deserialize to the same field element). perhaps: 

```
let limbs = [
    u64::from_be_bytes(bytes[ 0.. 8].try_into().unwrap()),
    u64::from_be_bytes(bytes[ 8..16].try_into().unwrap()),
    u64::from_be_bytes(bytes[16..24].try_into().unwrap()),
    u64::from_be_bytes(bytes[24..32].try_into().unwrap()),
];
let bigint = BigInteger256::new(limbs);

if bigint >= Fq::MODULUS {
    return Err(D::Error::custom("value not in canonical field range"));
}

Ok(Self(Fq::from_bigint(bigint).ok_or_else(|| D::Error::custom("invalid field element"))?))
```

or

```
let mut mod_be = [0u8; 32];
{
    let m = Fq::MODULUS.to_bytes_be();
    mod_be[32 - m.len()..].copy_from_slice(&m);
}

if bytes >= &mod_be {
    return Err(D::Error::custom("value not in canonical field range"));
}

Ok(Self(Fq::from_be_bytes_mod_order(bytes)))
```

crates/world-id-core/src/credential.rs
```diff
@@ -0,0 +1,157 @@
...
+    ) -> Result<Self, anyhow::Error> {
+        self.associated_data_hash = associated_data_hash.try_into()?;
+        Ok(self)
+    }
+
+    /// Get the credential domain separator for the given version.
+    pub fn get_cred_ds(&self) -> BaseField {
+        match self.version {
+            CredentialVersion::V1 => {
+                BaseField::from_be_bytes_mod_order(b"POSEIDON2+EDDSA-BJJ+DLBE-v1")
```
decentralgabe, line 118:
```suggestion
                BaseField::from_be_bytes_mod_order(b"POSEIDON2+EDDSA-BJJ")
```

not sure if want to add +CBOR at the end

decentralgabe (approved):
lgtm pending one comment

PR #19, opened 2025-09-29
feat: issuer schema registry

paolodamico:
- Renames the issuer contract to `IssuerSchema` to reflect the specific issuer+schema pairs that will be registered in this contract.
- Allow registering a schemaUri for each issuer schema.


decentralgabe (dismissed):
lgtm

philsippl:
for what do we have then the `issuerId` registered actually? Should we rename this to the `issuerSchemaId` and remove the rest again?

PR #24, opened 2025-10-02
feat: housekeeping and dictator clippy

paolodamico:
Some clean up, refactoring & documentation. Started feature flagging `authenticator` as well

### DIFF
--- a/crates/world-id-core/src/authenticator.rs
+++ b/crates/world-id-core/src/authenticator.rs
@@ -58,9 +58,9 @@ type MerkleProof = (BaseField, [BaseField; TREE_DEPTH], MerkleEpoch);
 /// An Authenticator is the base layer with which a user interacts with the Protocol.
 #[derive(Debug)]
 pub struct Authenticator {
-    signer: AuthenticatorSigner,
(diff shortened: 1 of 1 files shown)


crates/world-id-core/src/authenticator.rs
```diff
@@ -237,3 +302,49 @@ impl Authenticator {
...
+
+impl AuthenticatorSigner {
+    /// Initializes a new signer from an input seed.
+    pub fn from_seed_bytes(seed: &[u8]) -> Result<Self> {
+        if seed.len() != 32 {
+            return Err(eyre::eyre!("seed must be 32 bytes"));
+        }
+        let bytes: [u8; 32] = seed.try_into()?;
+        let onchain_signer = PrivateKeySigner::from_bytes(&bytes.into())?;
+        let offchain_signer = EdDSAPrivateKey::from_bytes(bytes);
```
decentralgabe, line 325: is there any risk in reusing a seed for two different curves?

decentralgabe: i guess one seed compromise compromises both, but then we have to store multiple seeds

PR #53, opened 2025-10-15
feat: add issuer support

paolodamico:
Adds utilities for the Issuer and ships with an integration test all the way to the contract.


paolodamico:
@codex review

PR #54, opened 2025-10-16
feat: flip signer <-> id lookup for cred registry

paolodamico:
Each schemaIssuerId must have one signer, but a single signer can register many schemas.


paolodamico:
@codex review

PR #64, opened 2025-10-21
feat: introduce generic types crate

paolodamico:
The lightweight `world-id-types` crate will only define raw types to use across all other crates and services. This will help standardize shared types between the services in this monorepo and https://github.com/TaceoLabs/nullifier-oracle-service.

Particularly this PR:
1. Imports the `rpId` type from nullifier-oracle-service monorepo.
2. Introduces a custom `WorldIdProof` type that merges the groth16 proof with the merkle root.
3. Introduces a custom `FieldElement` with serialization specific for the Protocol.

### DIFF
--- a/crates/types/src/lib.rs
+++ b/crates/types/src/lib.rs
@@ -12,7 +12,7 @@
 )]
 
 use ark_babyjubjub::Fq;
-use ark_ff::{AdditiveGroup, Field};
+use ark_ff::{AdditiveGroup, Field, PrimeField};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ruint::aliases::U256;
 use serde::{de::Error as _, ser::Error as _, Deserialize, Deserializer, Serialize, Serializer};
@@ -45,6 +45,7 @@ impl FieldElement {
     pub const ZERO: Self = Self(Fq::ZERO);
     /// The multiplicative identity of the field.
     pub const ONE: Self = Self(Fq::ONE);
+
     /// Serializes the field element into a byte vector.
     ///
     /// # Errors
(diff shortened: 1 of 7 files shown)


crates/types/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+/// defined for the curve that is used to sign credentials in the World ID Protocol.
+///
+/// The World ID Protocol currently uses the `BabyJubJub` curve.
+pub type BaseField = ark_babyjubjub::Fq;
+
+/// Represents a Merkle root hash for any of the trees in the World ID Protocol.
+///
+/// The inner type is a base field element from `BabyJubJub` for convenience instead of a scalar field element on BN254.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+pub struct FieldElement(BaseField);
```
philsippl: I find it a bit confusing to have this wrapper for a wrapper. Maybe let's make `BaseField` already a custom type and do something like this?

```rust
pub struct BaseField(pub ark_babyjubjub::Fq);

impl Deref for BaseField {
    type Target = ark_babyjubjub::Fq;
    fn deref(&self) -> &Self::Target { &self.0 }
}
impl DerefMut for BaseField {
    fn deref_mut(&mut self) -> &mut Self::Target { &mut self.0 }
}
```

crates/types/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+impl From<BaseField> for FieldElement {
+    fn from(value: BaseField) -> Self {
+        Self(value)
+    }
+}
+
+impl fmt::Display for FieldElement {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let u256: U256 = (*self).into();
+        write!(f, "{u256:#066x}")
```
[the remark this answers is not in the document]
philsippl: is the custom formatter actually the only reason for the `FieldElement` type? (I don't see a good reason actually?)

crates/types/src/rp.rs
```diff
@@ -0,0 +1,162 @@
...
+
+use std::{fmt, str::FromStr};
+
+use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::BaseField;
+
+/// The id of a relying party.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct RpId(u128);
```
philsippl: is there a reason we go with u128 here?

PR #84, opened 2025-10-29
feat: import oprf-client, oprf-world-types & oprf-zk

paolodamico:
- Imports the `oprf-client`, `oprf-world-types` & `oprf-zk` crates from https://github.com/TaceoLabs/nullifier-oracle-service with verbatim logic.
- Updates existing crates to use the newly imported crates.
- Standardizes some of the pother crate dependencies to ensure all crates use the same deps.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ ark-bn254 = "0.5"
 ark-ff = "0.5"
 ark-groth16 = "0.5"
 ark-serialize = "0.5"
-circom-witness-rs = "0.2.0"
+circom-witness-rs = "0.2.1"
 circom-types = { git = "https://github.com/TaceoLabs/co-snarks.git", default-features = false }
 criterion = "0.7"
 eddsa-babyjubjub = { package = "taceo-eddsa-babyjubjub", version = "0.5" }
--- a/crates/core/bin/rp.rs
+++ b/crates/core/bin/rp.rs
@@ -3,15 +3,14 @@ use std::time::{Duration, SystemTime};
 use alloy::{
     primitives::{address, Address},
     providers::ProviderBuilder,
-    signers::k256::{self, ecdsa::signature::SignerMut},
+    signers::k256::{self, ecdsa::signature::SignerMut, SecretKey},
 };
 use ark_ff::BigInteger;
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -12,26 +15,48 @@ resolver = "2"
...
+ark-ec = "0.5"
+ark-babyjubjub = { package = "taceo-ark-babyjubjub", version = "0.5" }
+ark-serde-compat = { package = "taceo-ark-serde-compat", version = "0.1" }
 ark-bn254 = "0.5"
 ark-ff = "0.5"
 ark-groth16 = "0.5"
 ark-serialize = "0.5"
-eddsa-babyjubjub = { git = "https://github.com/TaceoLabs/nullifier-oracle-service", rev = "698f03f" }
-hex = {version = "0.4" }
+circom-witness-rs = "0.2.0"
```
fabian1409: in order for us to be able to use orpf-zk in our repo (which we will need to do unless we generalize it more and move it to a extra crate) we need the latest git version of circom-witness-rs

crates/core/bin/rp.rs
```diff
@@ -1,34 +1,39 @@
...
     let mut rng = rand::thread_rng();
     let chain_url = std::env::var("CHAIN_URL").unwrap_or("ws://localhost:8545".to_string());
     let action_id: FieldElement = ark_babyjubjub::Fq::rand(&mut rng).into();
 
-    let wallet = EthereumWallet::from(PRIVATE_KEY.parse::<PrivateKeySigner>()?);
-    let mut key_gen_proxy =
-        KeyGenProxy::connect(&chain_url, DEFAULT_KEY_GEN_CONTRACT_ADDRESS, wallet).await?;
-
-    let (rp_id, rp_nullifier_key) = key_gen_proxy.init_key_gen().await?;
+    let rp_pk = EcDsaPubkeyCompressed::try_from(MOCK_RP_SECRET_KEY.public_key())?;
```
fabian1409: keep in mind that this a hardcoded secret key, only intended for e2e tests

lukejmann (approved):

Cargo.toml
paolodamico: version bumped

PR #89, opened 2025-11-05
Allow re-using authenticator address after an old account recovers

karankurbur:
Description: The AccountRegistry contract allows users to create accounts and register one or more authenticator addresses for each ac-
count. To prevent an authenticator address from being registered to multiple accounts, the _registerAccount(...) and insertAuthenticator(...)
functions check if the address is already present in the authenticatorAddressToPackedAccountIndex mapping.

When creating an account with a previously used authenticator address, we check if the recovery counter is less than the accountIndex's recovery counter. If so, it means that the user has recovered their account and the authenticator address can be re-used by a different user.

### DIFF
--- a/contracts/out/AccountRegistry.sol/AccountRegistryAbi.json
+++ b/contracts/out/AccountRegistry.sol/AccountRegistryAbi.json
@@ -1114,10 +1114,10 @@
   },
   {
     "type": "error",
-    "name": "AuthenticatorAddressAlreadyInUse",
+    "name": "AddressEmptyCode",
     "inputs": [
       {
-        "name": "authenticatorAddress",
+        "name": "target",
         "type": "address",
         "internalType": "address"
       }
@@ -1125,10 +1125,10 @@
   },
   {
     "type": "error",
-    "name": "AddressEmptyCode",
+    "name": "AuthenticatorAddressAlreadyInUse",
     "inputs": [
       {
-        "name": "target",
+        "name": "authenticatorAddress",
         "type": "address",
         "internalType": "address"
       }


contracts/src/lib/PackedAccountIndex.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+/**
```
paolodamico: you'll probably need to include this library in the anvil test deployments for the integration tests

contracts/src/AccountRegistry.sol
```diff
@@ -239,10 +243,21 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
+
+            uint256 packedAccountIndex = authenticatorAddressToPackedAccountIndex[authenticatorAddress];
+            // If the authenticatorAddress is non-zero, we could permit it to be used if the recovery counter is less than the
+            // accountIndex's recovery counter. This means the account was recovered and the authenticator address is not used.
+            if (packedAccountIndex != 0) {
+                uint256 existingAccountIndex = PackedAccountIndex.accountIndex(packedAccountIndex);
+                uint256 existingRecoveryCounter = PackedAccountIndex.recoveryCounter(packedAccountIndex);
+                require(
+                    existingRecoveryCounter < accountRecoveryCounter[existingAccountIndex],
+                    "Authenticator already exists"
```
paolodamico: probably also custom error?

paolodamico (dismissed):

paolodamico (dismissed):

crates/test-utils/src/anvil.rs
```diff
@@ -172,16 +181,47 @@ impl TestAnvil {
...
                 .context("failed to deploy BinaryIMT library")?;
 
-        // Step 3: Link BinaryIMT and deploy AccountRegistry
-        let account_registry_bytecode = Self::link_library(
-            include_str!(concat!(
-                env!("CARGO_MANIFEST_DIR"),
-                "/../../contracts/out/AccountRegistry.sol/AccountRegistry.json"
-            )),
+        // Step 3: Deploy PackedAccountIndex library (no dependencies)
+        let packed_account_index = PackedAccountIndex::deploy(provider.clone())
```
karankurbur: this works?

paolodamico (approved):

PR #90, opened 2025-11-05
add merkle sanity check loop

lukejmann:


services/indexer/src/lib.rs
```diff
@@ -307,6 +307,57 @@ async fn start_http_server(addr: SocketAddr, pool: PgPool) -> anyhow::Result<()>
     Ok(())
 }
 
+/// Periodically checks that the local in-memory Merkle root remains valid on-chain.
+async fn root_sanity_check_loop(
```
paolodamico: nit. let's move this to its own module to avoid fattening this?

paolodamico (approved):

services/indexer/src/lib.rs
lukejmann: updated

paolodamico (approved):

PR #96, opened 2025-11-09
fix: unify pubkey empty validation

paolodamico:
Previously we had two different definitions for a valid pubkey. This unifies and enforces both coordinates are non-zero.


paolodamico:
@codex review

PR #111, opened 2025-11-18
disambiguate leaf_index and account_index

paolodamico:
Drops the distinction between the two to only use `account_index` and for it to represent the logical index at the Merkle tree. This is done by inserting an empty leaf at 0-index when the tree is initialized.


paolodamico:
@codex review

PR #114, opened 2025-11-20
make pubkeyId & recoveryCounter explicitly uint32

paolodamico:
Especially for outside calls, making explicit that these are 32 bits.

### DIFF
--- a/contracts/src/AccountRegistry.sol
+++ b/contracts/src/AccountRegistry.sol
@@ -246,7 +246,7 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
 
     /**
      * @dev Helper function to set pubkey bitmap packed, preserving the recovery address. The
-     * bitmap is 32 bits, but 256 are accepted to simplify bit operations in other functions.
+     * bitmap is 96 bits, but 256 are accepted to simplify bit operations in other functions.
      */
     function _setPubkeyBitmap(uint256 accountIndex, uint256 bitmap) internal {
         if (bitmap >> 96 != 0) {
@@ -261,7 +261,7 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
 
     /**
      * @dev Helper function to set recovery address and pubkey bitmap packed. The
-     * bitmap is 32 bits, but 256 are accepted to simplify bit operations in other functions.
+     * bitmap is 96 bits, but 256 are accepted to simplify bit operations in other functions.
      */
     function _setRecoveryAddressAndBitmap(uint256 accountIndex, address recoveryAddress, uint256 bitmap) internal {
         if (bitmap >> 96 != 0) {


contracts/src/AccountRegistry.sol
```diff
@@ -218,27 +232,36 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
     }
 
     /**
-     * @dev Helper function to get pubkey bitmap from packed
+     * @dev Helper function to get pubkey bitmap from the packed storage.
      */
     function _getPubkeyBitmap(uint256 accountIndex) internal view returns (uint256) {
-        return _accountIndexToRecoveryAddressPacked[accountIndex] >> 160;
+        return uint32(_accountIndexToRecoveryAddressPacked[accountIndex] >> 224);
```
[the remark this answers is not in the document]
paolodamico: the contract will be redeployed, not relevant

contracts/src/AccountRegistry.sol
```diff
@@ -34,7 +25,7 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
     //                        Members                         //
     ////////////////////////////////////////////////////////////
 
-    // accountIndex -> [96 bits bitmap of pubkeyIds][160 bits recoveryAddress]
+    // accountIndex -> [32 bits pubkeyId bitmap][64 bits unused][160 bits recoveryAddress]
```
philsippl: Is there maybe a confusion here of pubkeyIds and the bitmap of pubkeys or why reducing this? This also has the effect that there can only be 32 pubkeys registered.

contracts/src/AccountRegistry.sol
```diff
@@ -218,27 +232,36 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
+        return uint32(_accountIndexToRecoveryAddressPacked[accountIndex] >> 224);
     }
 
     /**
-     * @dev Helper function to set pubkey bitmap in packed
+     * @dev Helper function to set pubkey bitmap packed, preserving the recovery address. The
+     * bitmap is 32 bits, but 256 are accepted to simplify bit operations in other functions.
      */
     function _setPubkeyBitmap(uint256 accountIndex, uint256 bitmap) internal {
+        if (bitmap >> 224 != 0) {
```
philsippl: Don't you want to check here that only the lowest 32 bit are set? i.e. `bitmap >> 32 != 0`

contracts/src/AccountRegistry.sol
```diff
@@ -218,27 +232,36 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
+            revert BitmapOverflow();
+        }
+
         uint256 packed = _accountIndexToRecoveryAddressPacked[accountIndex];
         // Clear bitmap bits and set new bitmap
-        packed = (packed & uint256(type(uint160).max)) | (bitmap << 160);
+        packed = (packed & uint256(type(uint160).max)) | (uint256(bitmap) << 224);
```
philsippl: unnecessary cast

PR #124, opened 2025-11-22
feat: world-id crate for publishing

paolodamico:
Adds a `world-id` crate to wrap `world-id-core`. This is for future proofing when more crates are introduced.


paolodamico:
@codex review

PR #125, opened 2025-11-24
feat: RPC is now optional for the authenticator

paolodamico:
Makes relevant changes for the RPC to be optional for the authenticator.

1. Introduce an endpoint in the indexer to fetch the signing nonce from on-chain for an account.
2. Standardize indexer responses (incl. errors) between the core crate and indexer.
3. RPC URL is now optional. If not provided, the Authenticator will fallback to the indexer.


paolodamico:
@codex review

PR #126, opened 2025-11-24
feat: publishing for world-id-core

paolodamico:
Adds required build script to publish `world-id-core` to crates.io. The zkeys must be fetched remotely at build-time because they exceed the max size allowed to be submitted to crates.io


paolodamico:
@codex review

PR #130, opened 2025-12-01
Use custom errors and rename lib/mappings

karankurbur:
Copying over changes from https://github.com/worldcoin/world-id-protocol/pull/92 instead of rebasing

1. Rename vars to keep consistency and readability
authenticatorAddressToPackedAccountIndex -> authenticatorAddressToPackedAccountData
accountIndexToSignatureNonce
accountIndexToRecoveryCounter

2. Rename lib to PackedAccountData

3. Modify helper function -> recoverAccountDataFromSignature to return signer and packedAccountData

4. Use custom errors

5. Remove zero address check on _registerAccount - this is optional according to the spec

### DIFF
--- a/contracts/src/AccountRegistry.sol
+++ b/contracts/src/AccountRegistry.sol
@@ -424,7 +424,7 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
         }
         packedAccountData = authenticatorAddressToPackedAccountData[signer];
         if (packedAccountData == 0) {
-            revert AccountDoesNotExist(0);
+            revert AuthenticatorDoesNotExist(signer);
         }
         uint256 accountIndex = PackedAccountData.accountIndex(packedAccountData);
         uint256 actualRecoveryCounter = PackedAccountData.recoveryCounter(packedAccountData);


paolodamico:
@codex review

contracts/src/AccountRegistry.sol
```diff
@@ -294,38 +411,48 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
-        packedAccountIndex = authenticatorAddressToPackedAccountIndex[signer];
-        require(packedAccountIndex != 0, "Account does not exist");
-        accountIndex = uint256(uint192(packedAccountIndex));
-        require(packedAccountIndex >> 224 == accountRecoveryCounter[accountIndex], "Invalid account recovery counter");
+        if (signer == address(0)) {
+            revert ZeroRecoveredSignatureAddress();
+        }
+        packedAccountData = authenticatorAddressToPackedAccountData[signer];
+        if (packedAccountData == 0) {
+            revert AccountDoesNotExist(0);
```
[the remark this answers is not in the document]
paolodamico: not applicable

contracts/src/AccountRegistry.sol
```diff
@@ -442,13 +571,20 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
         uint256[] calldata siblingNodes,
         uint256 nonce
     ) external virtual onlyProxy onlyInitialized {
-        _validateAuthenticatorAddressNotInUse(newAuthenticatorAddress);
-        require(
-            oldAuthenticatorAddress != newAuthenticatorAddress, "Old and new authenticator addresses cannot be the same"
-        );
-        require(newAuthenticatorAddress != address(0), "New authenticator address cannot be the zero address");
+        if (accountIndex == 0 || nextAccountIndex <= accountIndex) {
```
paolodamico: is this extra gas worth it?

contracts/src/AccountRegistry.sol
```diff
@@ -658,14 +836,20 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
-        require(SignatureChecker.isValidSignatureNow(recoverySigner, messageHash, signature), "Invalid signature");
-        require(authenticatorAddressToPackedAccountIndex[newAuthenticatorAddress] == 0, "Authenticator already exists");
-        require(newAuthenticatorAddress != address(0), "New authenticator address cannot be the zero address");
+        if (!SignatureChecker.isValidSignatureNow(recoverySigner, messageHash, signature)) {
+            revert InvalidSignature();
+        }
+        if (authenticatorAddressToPackedAccountData[newAuthenticatorAddress] != 0) {
+            revert AuthenticatorAlreadyExists(newAuthenticatorAddress);
+        }
+        if (newAuthenticatorAddress == address(0)) {
```
paolodamico: nit. maybe at the top to save some gas?

paolodamico (dismissed):

contracts/src/AccountRegistry.sol
karankurbur: leaving as is, not in common flows (only in update and recover functions)

contracts/src/AccountRegistry.sol
```diff
@@ -535,13 +683,21 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
+        uint256 recoveredAccountIndex = PackedAccountData.accountIndex(packedAccountData);
+        if (accountIndex != recoveredAccountIndex) {
+            revert MismatchedAccountIndex(accountIndex, recoveredAccountIndex);
+        }
+
+        uint256 expectedNonce = accountIndexToSignatureNonce[accountIndex];
+        if (nonce != expectedNonce) {
+            revert MismatchedSignatureNonce(accountIndex, expectedNonce, nonce);
+        }
+        accountIndexToSignatureNonce[accountIndex]++;
```
paolodamico: **blocking**. actually this will be problematic. for just this function we're increasing the nonce after the check, but for every other function we're increasing it before the check. I think after the check is more natural, but we should update it to have the same behavior everywhere. cc @philsippl

contracts/src/AccountRegistry.sol
```diff
@@ -294,38 +409,48 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
-        packedAccountIndex = authenticatorAddressToPackedAccountIndex[signer];
-        require(packedAccountIndex != 0, "Account does not exist");
-        accountIndex = uint256(uint192(packedAccountIndex));
-        require(packedAccountIndex >> 224 == accountRecoveryCounter[accountIndex], "Invalid account recovery counter");
+        if (signer == address(0)) {
+            revert ZeroRecoveredSignatureAddress();
+        }
+        packedAccountData = authenticatorAddressToPackedAccountData[signer];
+        if (packedAccountData == 0) {
+            revert AccountDoesNotExist(0);
```
paolodamico: actually I think copilot was right, what doesn't exist here is the authenticator address, `AuthenticatorDoesNotExist` would seem like a better error

PR #131, opened 2025-12-03
feat: account registry contract improvements

paolodamico:
1. Change the onlyInitialized modifier to reduce code size (linter suggestion)
2. Change `maxAuthenticators` to a variable that can be updated by the contract owner.
3. Update function name `_recoverAccountDataFromSignature` to follow underscore convention for internal functions.
4. Ensure a zero address authenticator cannot be added when recovering an account.

### DIFF
--- a/contracts/src/AccountRegistry.sol
+++ b/contracts/src/AccountRegistry.sol
@@ -267,6 +267,11 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
      */
     error AuthenticatorDoesNotBelongToAccount(uint256 expectedAccountIndex, uint256 actualAccountIndex);
 
+    /**
+     * @dev Thrown when trying to update max authenticators beyond the natural limit.
+     */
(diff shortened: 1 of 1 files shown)


contracts/src/AccountRegistry.sol
```diff
@@ -903,6 +900,28 @@ contract AccountRegistry is Initializable, EIP712Upgradeable, Ownable2StepUpgrad
...
+        emit RootValidityWindowUpdated(old, newWindow);
+    }
+
+    /**
+     * @dev Set an updated maximum number of authenticators allowed.
+     */
+    function setMaxAuthenticators(uint256 newMaxAuthenticators) external onlyOwner onlyProxy onlyInitialized {
+        uint256 old = maxAuthenticators;
+        maxAuthenticators = newMaxAuthenticators;
+        emit MaxAuthenticatorsUpdated(old, maxAuthenticators);
```
karankurbur: do we need to enforce a max value on maxAuthenticators since we use are using a bitmap

PR #150, opened 2025-12-08
feat: poseidon2 hash sponge construction from raw bytes

carlomazzaferro:
As internally discussed, we'd like to just allow passing raw bytes to instantiate it the `Credential` allowing the protocol to take care of hashing and mapping to the `FieldElement`. This comes from the realization that simply just hashing using e.g. sha256 or blake3, and passing the hash might fail due to the req of the FE having 254 bit prime modulus.

---

> [!NOTE]
> Introduce a Poseidon2-based SAFE-style sponge for hashing arbitrary bytes to field elements and expose `Credential::associated_data`/helper to use it, with tests and `sha3` dependency.
> 
> - **Primitives/Credential**:
>   - Add `Credential::associated_data(&[u8])` to set `associated_data_hash` from raw bytes via Poseidon2.
>   - Add `Credential::hash_bytes_to_field_element(&[u8])` wrapper (SAFE-style, t=16, r=15) using `sponge::hash_bytes_with_poseidon2_t16_r15` and a domain separator `DS_TAG`.
>   - Extend tests to cover determinism, error cases, chunking/batching, and parity with `associated_data`.
> - **Sponge Module** (`crates/primitives/src/sponge.rs`):
>   - New SAFE-style sponge utilities: `IoPattern`, `derive_safe_tag`, `bytes_to_field_elements`, and `hash_bytes_with_poseidon2_t16_r15` (Poseidon2 t=16; SHA3-256 tag in capacity).
>   - Unit tests for tag stability, empty input rejection, and determinism.
> - **Library**:
>   - Expose `pub mod sponge` in `lib.rs`.
> - **Dependencies**:
>   - Add `sha3` for tag derivation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b1fbb80f7c27f16167b194301644145c9ff0b89. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/crates/primitives/src/credential.rs
+++ b/crates/primitives/src/credential.rs
@@ -187,8 +187,7 @@ impl Credential {
     /// 1. Compute a SAFE-style tag from an IO pattern that encodes the input
     ///    length (in bytes), the squeeze size (32 bytes), and a domain separator.
     ///    The tag is derived by hashing these bytes with SHA3-256 and reducing to
-    ///    a field element (SAFE would normally place this in capacity; we place it
-    ///    in the rate per the design note for this usage).
+    ///    a field element (placed in the capacity element, per SAFE guidance).
     /// 2. Split input into 31-byte chunks, convert each to a field element.
     /// 3. Absorb at most 15 field elements at a time (add into rate), then
     ///    permute (Poseidon2 t16) after each batch.
--- a/crates/primitives/src/sponge.rs
+++ b/crates/primitives/src/sponge.rs
@@ -108,7 +108,7 @@ pub fn bytes_to_field_elements(chunk_size: usize, data: &[u8]) -> Vec<Fq> {
 }
 
 /// Hash arbitrary bytes to a field element with Poseidon2 (t=16, rate=15, capacity=1),
-/// using a SAFE-style tag placed in the rate portion (per design note).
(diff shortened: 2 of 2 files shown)


paolodamico (approved):
lgtm but @dkales or @philsippl could you take a look as well?

crates/primitives/src/credential.rs
```diff
@@ -162,6 +164,96 @@ impl Credential {
...
+                state[i] += elem;
+            }
+            // Apply permutation after each batch
+            poseidon2.permutation_in_place(&mut state);
+        }
+
+        // Padding marks the end of data absorption; prefix-freedom is ensured by the combination
+        // of this padding and the length encoding below, which prevents collisions between inputs
+        // where one is a prefix of another.
+        state[0] += Fq::ONE;
```
dkales: This method allows for length extension attacks, since field elements that are not exact multiples of the rate get implicitly padded with zeros above. This means that something like [123]  and [123,0,0,0,0,0,0,0] (in FE) would produce the same hash, which is not wanted.
My suggestion would be to implement the SAFE API for this: https://eprint.iacr.org/2023/522, this handles the lengths and padding as well.

dkales (changes requested):

PR #155, opened 2025-12-10
feat: gateway multicall parse

lukejmann:
> [!NOTE]
> Adds per-subcall status handling for Multicall3 using debug trace, switches to allowFailure=true with buffered gas estimation, and introduces an integration test for partial failures.
> 
> - **Gateway (ops batching)**:
>   - Switch `Multicall3::Call3.allowFailure` to `true` and send aggregated ops via `aggregate3`.
>   - Add gas estimation using cloned calls with `allowFailure=false`; apply 20% buffer and set `gas(...)` on the tx; fall back to default gas if estimation fails.
>   - On successful tx receipt, invoke `debug_traceTransaction` (`callTracer`) to collect internal calls to `AccountRegistry` and:
>     - Mark each corresponding request `Finalized` or `Failed` based on subcall success; parse revert data via `parse_contract_error`.
>     - Fallback: if trace unsupported or counts mismatch, mark all as `Finalized`.
>   - Preserve existing batch status updates and error handling; import `Provider` for debug RPC access.
> - **Tests**:
>   - Add `test_multicall_partial_failure_debug_trace` to assert batched ops where one subcall succeeds and one fails, with correct per-request statuses and on-chain state.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30f064087f04d7b6749bd395f13e9476b3c14857. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7361,6 +7361,7 @@ dependencies = [
  "ruint",
  "serde",
  "serde_json",
+ "sha3",
  "taceo-ark-babyjubjub",
  "taceo-ark-serde-compat",
  "taceo-circom-types",
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -34,6 +34,7 @@ thiserror = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 k256 = { workspace = true, features = ["ecdsa", "serde"] }
 poseidon2 = { workspace = true }
+sha3 = { version = "0.10", default-features = false }
 
 [dev-dependencies]
 ciborium = "0.2.2"
--- a/crates/primitives/src/credential.rs
+++ b/crates/primitives/src/credential.rs
@@ -3,7 +3,10 @@ use eddsa_babyjubjub::{EdDSAPublicKey, EdDSASignature};
 use ruint::aliases::U256;
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 
-use crate::{FieldElement, PrimitiveError};
+use crate::{sponge::hash_bytes_with_poseidon2_t16_r15, FieldElement, PrimitiveError};
+
+/// Domain separation tag to avoid collisions with other Poseidon2 usages.
+const DS_TAG: &[u8] = b"ASSOCIATED_DATA_HASH_V1";
 
 /// Version of the `Credential` object
 #[derive(Default, Debug, PartialEq, Eq, Hash, Copy, Clone, Serialize, Deserialize)]
@@ -148,7 +151,7 @@ impl Credential {
         Ok(self)
     }
 
-    /// Set the associated data hash of the credential.
+    /// Set the associated data hash of the credential from a pre-computed hash.
     ///
     /// # Errors
     /// Will error if the provided hash cannot be lowered into the field.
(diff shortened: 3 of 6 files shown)


services/gateway/src/ops_batcher.rs
```diff
@@ -262,18 +283,156 @@ impl OpsBatcherRunner {
...
-                                    tracker
-                                        .set_status_batch(
-                                            &ids_for_receipt,
-                                            RequestState::Finalized {
-                                                tx_hash: hash.clone(),
-                                            },
+                                    // Transaction succeeded. Since we set allowFailure=true in Multicall3,
+                                    // we need to inspect which subcalls (to the registry) failed.
+                                    // Try to debug_trace the transaction; if this fails (e.g. provider
+                                    // does not support debug), fall back to marking all as finalized.
```
karankurbur: can we check if the provider supports debug trace before using `allowFailure: true`? 

We should avoid at all costs the possibility of setting something to Finalized that actually failed.

services/gateway/src/ops_batcher.rs
```diff
@@ -262,18 +283,156 @@ impl OpsBatcherRunner {
...
+                                                }
+                                            }
+
+                                            let target_hex = format!("0x{:x}", registry);
+                                            let mut subcalls: Vec<&serde_json::Value> = Vec::new();
+                                            let root = trace.get("result").unwrap_or(&trace);
+                                            collect_calls(root, &target_hex, &mut subcalls);
+
+                                            // If counts mismatch, fall back to finalize-all to avoid mislabeling.
+                                            if subcalls.len() != ids_for_receipt.len() {
```
karankurbur: This should trigger an alert to investigate. This should never happen

Lets throw an error and put in error state.

services/gateway/src/ops_batcher.rs
```diff
@@ -262,18 +283,156 @@ impl OpsBatcherRunner {
...
+                                                    let failed_flag = sc
+                                                        .get("reverted")
+                                                        .and_then(|b| b.as_bool())
+                                                        .unwrap_or(false)
+                                                        || sc
+                                                            .get("failed")
+                                                            .and_then(|b| b.as_bool())
+                                                            .unwrap_or(false);
+                                                    // Try to extract revert data for better error code parsing.
+                                                    let revert_data = sc
```
karankurbur: GPT says that output is always `0x` when it reverts so should always check `revertReason`

services/gateway/src/ops_batcher.rs
```diff
@@ -262,18 +283,156 @@ impl OpsBatcherRunner {
...
-                                                tx_hash: hash.clone(),
-                                            },
+                                    // Transaction succeeded. Since we set allowFailure=true in Multicall3,
+                                    // we need to inspect which subcalls (to the registry) failed.
+                                    // Try to debug_trace the transaction; if this fails (e.g. provider
+                                    // does not support debug), fall back to marking all as finalized.
+                                    let traced = provider
+                                        .client()
+                                        .request::<_, serde_json::Value>(
+                                            "debug_traceTransaction",
```
aurel-fr: this is not a reliable pattern, the contract should emit logs and those should be parsed.

paolodamico:
after a sync today we decided to maintain `allowFailure = false` and fail the whole batch. creation operations are separately batched in the `createMany` call to the contract, which means the volume of operations expected for multicalls will be low. the complexity and cost of allowing inner call failures, parsing the whole traces, varying support from RPCs is not proportional to the gains.

instead we will do improvements #180 & #181

PR #171, opened 2026-01-12
chore(authenticators): use backon for retry

alessandromazza98:
This PR doesn't change the logic, keeping it as it's right now. I rewrote the retryability by using `backon` crate that makes it easier to read and understand that snippet of code.


paolodamico:
this will no longer be needed with #174 as we're delegating the retry logic to the actual authenticator provider. but very solid pattern, taking it for the tests. ty @alessandromazza98 🚀

PR #174, opened 2026-01-12
feat: delegate account creation polling

paolodamico:
### Changes
- Instead of exposing a blocking function that handles the asynchronous account creation process, we return an object that handles the pending state and delegate the polling logic to the authenticator provider. This is because particularly for mobile authenticators the threading logic is better managed by clients (cc @sideround).
  - This also enables authenticators to provide custom UI/UX when the process is taking a longer amount of time.
-  Bumps some dependencies minor versions for services.

---

> [!NOTE]
> Adds a non-blocking account registration flow and shifts polling responsibility to clients.
> 
> - Introduces `InitializingAuthenticator` returned by `Authenticator::register`, plus `init_or_register` convenience; removes prior `init_or_create(_blocking)` usage across bin/tests/clients
> - Uses `backon` exponential retry for polling; features enabled via `backon` workspace dep with `tokio-sleep`
> - Refines errors: `GatewayError`/`IndexerError` now use `reqwest::StatusCode`, adds `RegistrationError { error_code, error_message }`, and simplifies `Timeout`
> - Replaces `Authenticator::leaf_hash` helper with `AuthenticatorPublicKeySet::leaf_hash()` usages
> - Updates tests (registration, e2e flows, insert/update/remove) and `oprf-dev-client` to the new API; bumps minor deps (e.g., `alloy-node-bindings = "1"`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d616a407932bb7a96f392eeb206b89bdf493d09. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


0xOsiris (approved):

crates/core/bin/authenticator.rs
```diff
@@ -31,7 +35,60 @@ async fn main() -> Result<()> {
...
+
+                // Poll gateway until finalized, then retry init until indexer catches up (max 15s)
+                let start_time = std::time::Instant::now();
+                let max_duration = std::time::Duration::from_secs(15);
+
+                let poller = || async {
+                    // Check timeout
+                    if start_time.elapsed() > max_duration {
+                        return Err(eyre::eyre!("timeout after 15 seconds"));
+                    }
```
alessandromazza98: chore: if you want, `backon` has a feature for timeout as well. See `with_timeout` fn in the doc: https://docs.rs/backon/latest/backon/struct.Retry.html#method.with_timeout

crates/core/bin/authenticator.rs
```diff
@@ -31,7 +35,60 @@ async fn main() -> Result<()> {
...
+
+                poller
+                    .retry(
+                        ExponentialBuilder::default()
+                            .with_max_delay(std::time::Duration::from_secs(2)),
+                    )
+                    .when(|e| {
+                        // Only retry on transient errors, stop on permanent failures
+                        let msg = e.to_string();
+                        msg.contains("not yet synced") || msg.contains("not finalized")
```
alessandromazza98: for the future: string error messages can cause problems in the future if someone change the error message but it doesn't change here. In the future we can create an enum that contains all possible errors and then we can really match against the enum variants here

alessandromazza98 (dismissed):
left some suggestions. Still good to merge to me, we can iterate on it in follow up PRs

0xOsiris (dismissed):

Issue #175, opened 2026-01-12
Separate routes & improve module separation of gateway


paolodamico:
The gateway is primarily one big lib.rs file which makes it harder for new contributors to parse and contribute to. We should instead separate at least each route into its own file as we have for the indexer.

PR #178, opened 2026-01-13
refactor(gateway): improve modules separation

alessandromazza98:
Closes https://github.com/worldcoin/world-id-protocol/issues/175


services/gateway/src/routes/is_valid_root.rs
```diff
@@ -0,0 +1,39 @@
...
+    s.parse()
+        .map_err(|e| ApiError::bad_request(format!("invalid value: {}", e)))
+}
+
+pub(crate) async fn is_valid_root(
+    State(state): State<AppState>,
+    axum::extract::Query(q): axum::extract::Query<IsValidRootQuery>,
+) -> ApiResult<impl IntoResponse> {
+    let root = req_u256("root", &q.root)?;
+    let contract = WorldIdRegistry::new(state.registry_addr, state.provider.clone());
```
paolodamico: // TODO: we probably want to add a caching layer here

paolodamico (approved):

services/gateway/src/routes/is_valid_root.rs
alessandromazza98: Interesting, I'll think about it and eventually will create a related issue

PR #241, opened 2026-01-21
feat: Indexer - lower memory consumption

cichaczem:
> [!NOTE]
> Lowers memory usage and improves scalability for tree construction and cache maintenance.
> 
> - Replace `fetch_all_leaves` with cursor-based `fetch_leaves_batch(last_cursor, batch_size)` and add `fetch_and_parse_leaves_batch` for validated parsing
> - Rework `TreeBuilder::build_from_db_with_cache` to a two-pass chunked build: dense-prefix vector initialization, then sparse updates beyond the prefix; adds progress logging and returns last block/event IDs
> - Add unit test to verify chunk-based build matches traditional result
> - Refactor `TreeInitializer` API: `initialize()` now updates `GLOBAL_TREE` internally (no return value); split restore/replay and full rebuild into atomic `GLOBAL_TREE` updates and metadata writes; enhance `sync_with_db` to handle cache corruption via full rebuild
> - Update `lib.rs` to use new initializer contract and to log root from `GLOBAL_TREE`; cache refresh loop now calls `TreeInitializer::sync_with_db`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2074cb748dfc8fcd7cd502f7d8154eb3cde5975d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/services/indexer/src/tree/builder.rs
+++ b/services/indexer/src/tree/builder.rs
@@ -80,8 +80,10 @@ impl TreeBuilder {
         let mut last_cursor = String::from("0");
 
         loop {
-            let batch = self.fetch_and_parse_leaves_batch(pool, &last_cursor).await?;
-            
+            let batch = self
+                .fetch_and_parse_leaves_batch(pool, &last_cursor)
+                .await?;
+
             if batch.is_empty() {
                 break;
             }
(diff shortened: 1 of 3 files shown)


services/indexer/src/tree/builder.rs
```diff
@@ -50,29 +52,154 @@ impl TreeBuilder {
...
+        let mut total_leaves = 0u64;
+        let mut max_leaf_index = 0usize;
+        let mut last_cursor = String::from("0");
+
+        loop {
+            let batch = self
+                .fetch_and_parse_leaves_batch(pool, &last_cursor)
+                .await?;
+
+            if batch.is_empty() {
```
piohei: It may never be empty as new accounts are being added. ;)

PR #262, opened 2026-01-23
chore: CredentialSchemaIssuerRegistry refactor

karankurbur:
Making smaller PR

- Keep all interfaces
- Keep base files
- only update the code structure in CredentialSchemaIssuerRegistry


karankurbur:
Going to restart in a different PR instead of dealing with conflicts. Less chance of accidently changing logic and easier to keep it small

PR #289, opened 2026-01-28
chore: use interfaces for main contracts

karankurbur:
Define interfaces for the main contracts, put events, errors, and function definitions
Update event references to use Interface contract in both tests and rust code

No logic changes


Dzejkop (approved):
LGTM

PR #290, opened 2026-01-28
feat: extract credential and issuer into crates

0xForerunner:
This is the first PR of a few coming in that will extract different functionality into individual crates and get us out of this feature fiasco. Reasons for this are:

- Better compile times
- Easier to reason about when "features" are enabled and being compiled in at any moment
- Much Easier to get composability right, and making sure everything compiles for every possible combination of features.
- More idiomatic

My plan is to extract everything from world-id-core, but then keep world-id-core as a reexport crate which can export functionality from the smaller crates via feature flags. This way the API of world-id-core will not change for the user.


crates/issuer/src/signer.rs
```diff
@@ -0,0 +1,65 @@
+use alloy::{primitives::Address, signers::local::PrivateKeySigner};
+use eddsa_babyjubjub::{EdDSAPrivateKey, EdDSAPublicKey};
+use secrecy::{ExposeSecret, SecretBox};
+use world_id_primitives::PrimitiveError;
+
+/// The inner signer which can sign requests for both on-chain and off-chain operations. Both issuers and authenticators use this.
+///
+/// Both keys are zeroized on drop.
+#[derive(Debug)]
+pub struct Signer {
```
Dzejkop: Duplicate code exists in core/src/signer.rs

Dzejkop (approved):

PR #294, opened 2026-01-28
chore: use interface for VerifierNullifier

karankurbur:
- use interface for VerifierNullifier
- rename contract to match file name and update references
- Use MIT license in all files
- Use pragma solidity ^0.8.13 in all files

No logic changes


dkales:
It is generated using this: https://crates.io/crates/taceo-groth16-sol, 
which is based on the audited template by Remco in gnark, ensuring byte-for-byte identical output.

karankurbur:
There will be no change in the execution logic portion of the bytecode from either contract name changes or the usage of interfaces fyi. The full bytecode will be different as things like contract name are appended to the end of the executable bytecode

karankurbur:
Ok, i will leave this file untouched and rename the existing Verifier to WorldIDVerifier (the actual entry point)

PR #296, opened 2026-01-28
feat: create base class for inhertiance with upgradeable, 712, and fee logic

karankurbur:
In all four main contracts, we redefine several events, errors, modifiers, and methods that handle
- Upgradeability logic
- EIP712 logic
- Fee logic for registration

We create an abstract Base class that will be inherited by all that contains all the shared logic. 

This PR only updates the CredentialSchemaIssuer contract to use the Base class, separate PR's will address the other 3 contracts

### DIFF
--- a/contracts/src/abstract/WorldIDBase.sol
+++ b/contracts/src/abstract/WorldIDBase.sol
@@ -184,15 +184,4 @@ abstract contract WorldIDBase is Initializable, Ownable2StepUpgradeable, UUPSUpg
     /// @param newImplementation The address of the implementation being upgraded to.
     /// @custom:reverts string If called by any account other than the proxy owner.
     function _authorizeUpgrade(address newImplementation) internal virtual override onlyProxy onlyOwner {}
-
-    ////////////////////////////////////////////////////////////
-    //                       STORAGE GAP                      //
-    ////////////////////////////////////////////////////////////
-
-    /**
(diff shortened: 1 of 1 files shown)


contracts/src/abstract/Base.sol
```diff
@@ -0,0 +1,198 @@
+// SPDX-License-Identifier: MIT
```
karankurbur: [Reference impl](https://github.com/worldcoin/world-chain/blob/main/contracts/src/abstract/Base.sol)

contracts/src/abstract/Base.sol
```diff
@@ -0,0 +1,198 @@
...
+    ////////////////////////////////////////////////////////////
+    //                       STORAGE GAP                      //
+    ////////////////////////////////////////////////////////////
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[50] private __gap;
```
paolodamico: Given how we plan to do the updates, is this needed in the end?

karankurbur: We should clearly communicate this to auditors else they will def flag this

PR #304, opened 2026-01-29
chore(indexer): re-use multi transport provider

0xOsiris:
- Re-use global provider configuration in world-id-indexer. 
- Rename `common` crate to `world-id-common`


alessandromazza98 (dismissed):
fix CI, then good to merge. Approving to not block you further

PR #308, opened 2026-01-29
chore: add protocol to codeowners

0xForerunner:


murph (approved):
Not 100% sure this is a long term good idea, but for now yes

PR #367, opened 2026-02-05
feat: add session_nullifier

Takaros999:
This PR adds a `SesionNullifier` field which packs the nullifier + action which matches the `WorldIDVerifier.verifySession` function.

I tried making nullifier an enum at first but JSON serialization was akward. I ended up with 2 optional fields:
- `nullifier` for uniqueness proofs
- `session_nullifier` for session profs

I also added some checks in `.validate_response()` to avoid getting into a bad state.

### DIFF
--- a/crates/primitives/src/nullifier.rs
+++ b/crates/primitives/src/nullifier.rs
@@ -5,9 +5,6 @@ use crate::FieldElement;
 
 /// A session nullifier for World ID Session proofs.
 ///
-/// Session nullifiers intentionally allow RP-local linkability across interactions while
-/// preserving unlinkability across RPs.
-///
 /// They are an adaptation that reuses the same proof system inputs for session flows:
 /// - the nullifier component lets RPs detect replayed submissions for the same proof context
 /// - the action component is randomized for session verification semantics


crates/request/src/lib.rs
```diff
@@ -375,6 +431,30 @@ impl ProofRequest {
...
+            if self.session_id.is_some() {
+                // Session proof: must have session_nullifier
+                if response_item.session_nullifier.is_none() {
+                    return Err(ValidationError::MissingSessionNullifier(
+                        response_item.identifier.clone(),
+                    ));
+                }
+            } else {
+                // Uniqueness proof: must have nullifier
+                if response_item.nullifier.is_none() {
```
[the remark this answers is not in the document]
Takaros999: you can still initialize the struct on your own if you wanted too, I'm not sure if this needed

crates/primitives/src/nullifier.rs
```diff
@@ -0,0 +1,267 @@
...
+    #[must_use]
+    pub const fn action(&self) -> FieldElement {
+        self.action
+    }
+
+    /// Returns the session nullifier as an Ethereum-compatible array for `verifySession()`.
+    ///
+    /// Format: `[nullifier, action]` matching the contract's `uint256[2] sessionNullifier`.
+    #[must_use]
+    pub fn as_ethereum_representation(&self) -> [U256; 2] {
```
andy-t-wang:
```suggestion
    pub fn to_calldata(&self) -> [U256; 2] {
```
Perhaps this is a better name for the function

crates/primitives/src/nullifier.rs
```diff
@@ -0,0 +1,267 @@
...
+        let nullifier =
+            FieldElement::try_from(value[0]).map_err(|e| format!("invalid nullifier: {e}"))?;
+        let action =
+            FieldElement::try_from(value[1]).map_err(|e| format!("invalid action: {e}"))?;
+        Ok(Self { nullifier, action })
+    }
+
+    /// Converts to compressed bytes (64 bytes total: 2 x 32 bytes, big-endian).
+    #[must_use]
+    pub fn to_compressed_bytes(&self) -> Vec<u8> {
```
paolodamico: Nit. Given we're declaring the items in the struct as field elements and not numbers (U256), then let's encode using field elements, not numbers. you can do `FieldElement::serialize_as_bytes`

crates/primitives/src/nullifier.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+}
+
+impl Serialize for SessionNullifier {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if serializer.is_human_readable() {
```
paolodamico: for JSON, let's encode this as bytes and hex encode it. in fact, we may want to add a prefix for JSON that makes it explicit it's a session nullifier, e.g. `snil_`

crates/authenticator/src/authenticator.rs
```diff
@@ -561,7 +561,7 @@ impl Authenticator {
 
         let expires_at_min = request_item.effective_expires_at_min(request_timestamp);
 
-        let (proof, _public_inputs, nullifier) = generate_nullifier_proof(
+        let (proof, public_inputs, nullifier) = generate_nullifier_proof(
```
paolodamico: actually probably a better way to pull the action is to use `ProofRequest::computed_action`, and in fact we need to generate the random action there too.

paolodamico (approved):

PR #377, opened 2026-02-09
feat: Compute random action for session proofs

Takaros999:
This PR generates the action deterministically based on the request `nonce`. The idea behind this was to have the same action per request. 

Awaiting feedback from @paolodamico

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -307,10 +307,7 @@ jobs:
           targets: wasm32-unknown-unknown
 
       - name: Check primitives compiles for WASM
-        run: |
-          # Test WASM compilation without the 'circuits' feature (which has C++ deps)
-          # This ensures primitives can be used in WASM environments like browsers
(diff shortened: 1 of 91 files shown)


crates/request/src/lib.rs
```diff
@@ -347,12 +347,15 @@ impl ProofRequest {
...
+    /// When an explicit action is provided, it is returned directly.
+    /// For session proofs (action is `None`), a deterministic action is derived
+    /// from the request nonce via keccak256 hashing.
     #[must_use]
     pub fn computed_action(&self) -> FieldElement {
-        // if session_id -> compute differently
         match self.action {
             Some(action) => action,
-            None => todo!("Not ready"),
+            None => FieldElement::from_arbitrary_raw_bytes(self.nonce.to_string().as_bytes()),
```
paolodamico: hmm I'm not sure this is a good idea, this delegates the entropy to the RP and lets the RP somewhat influence the nullifier. it may be fine, but best to be on the safe side with a simple random field element generation here

PR #392, opened 2026-02-10
feat: use CascadingMerkleTree

0xForerunner:
`CascadingMerkleTree` is a fully parallelized and growable fixed depth merkle tree implementation. Tree capacity and memory consumption double as required when new leaves are pushed. We use Mmap backed Vecs as the underlying storage and support reloading from a file based cache.


0xOsiris:
Do we have benchmarks?

0xForerunner:
@0xOsiris can I get another approval?

0xOsiris (approved):

PR #396, opened 2026-02-11
refactor: oprf-node main wraped into run function for tracing clarity

0xThemis:
- **chore: log when watchers closed stream**
- **refactor: oprf-node main wraped into run function for tracing clarity**

### DIFF
--- a/contracts/foundry.lock
+++ b/contracts/foundry.lock
@@ -18,7 +18,7 @@
     }
   },
   "lib/oprf-key-registry": {
-    "rev": "0baafb8"
+    "rev": "2837db7"
   },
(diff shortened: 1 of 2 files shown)


services/oprf-node/src/auth/merkle_watcher.rs
```diff
@@ -187,6 +187,7 @@ impl MerkleWatcher {
                 loop {
                     let log = tokio::select! {
                         log = stream.next() => {
+                            tracing::warn!("MerkleWatcher subscribe stream was closed");
```
fabian1409: this warn would print always wouldnt it? could move it into the closure below

PR #398, opened 2026-02-11
ci: add World ID mobile ZK benchmarks

dcbuild3r:
Adds mobile benchmarking for the World ID ZK proof flows and wires it into BrowserStack-backed `mobench` CI. The branch is aligned with `mobench` / `mobench-sdk` `0.1.40`.

## Summary
- add the `crates/zk-mobile-bench` crate for World ID mobile proof benchmarks
- benchmark seven pi1/pi2 flows covering full proof generation, witness generation, proving-only, and cached query proof work
- expose UniFFI-compatible benchmark reports, including semantic phases and `cpu_median_ms` / `peak_memory_kb` resource usage fields
- add docs for local builds, BrowserStack runs, local native profiling, report summaries, and the `/mobench` trigger model
- update workspace metadata, lockfile, and AGENTS docs for the new crate

## CI / Workflow Changes
- add `mobile-bench.yml` as the manual and dispatched benchmark entrypoint
- install `mobench` from crates.io with `mobench_version: 0.1.40` and keep `mobench_ref` empty unless an explicit source override is needed
- add PR comment dispatch via `/mobench ...`, PR label auto-dispatch via the `bench` label, and automatic dispatch after the existing `Rust CI` / `Tests (stable)` check passes
- install mobile Rust targets only in the benchmark workflow jobs that need them
- install Noir in the iOS and Android benchmark jobs so proof dependencies are available during mobile benchmark builds
- preserve the local BrowserStack device-resolution patch path while allowing the workflow to skip it when the installed mobench already includes the fix
- render per-platform markdown summaries with plot support, publish SVG plot assets to `mobench-plots`, and update the sticky PR comment with GitHub-safe plot links
- upload raw BrowserStack outputs, recursive mobench result trees, rendered summaries, and history bundles for baseline comparison

## Trigger Model
- **PR comment**: `/mobench platform=both iterations=30 warmup=5`
- **PR label**: add the `bench` label; the auto workflow dispatches after `Rust CI` / `Tests (stable)` passes
- **Manual**: run the `Mobile Benchmarks` workflow from Actions

## Notes
- BrowserStack runs remain the canonical path for timing and memory benchmarks.
- BrowserStack native profiling is not supported by `mobench` `0.1.40`; local native profiling is documented through `cargo mobench profile run --provider local`.
- The reusable workflow still supports `mobench_ref` for source-ref overrides, but defaults to the crates.io `0.1.40` install path.


dcbuild3r:
Temporarily closing to retrigger mobile-bench checks after CI fix; reopening immediately.

dcbuild3r:
/mobench platform=both iterations=30 warmup=5 device_profile=low-spec

dcbuild3r:
/mobench platform=both iterations=30 warmup=5 device_profile=low-spec

rust-toolchain.toml
```diff
@@ -1,5 +1,13 @@
 [toolchain]
 channel = "1.93.0"
-targets = ["x86_64-unknown-linux-gnu"]
+targets = [
```
paolodamico: maybe best to just install the targets when required for running benchmarks? most contributors won't run benchmarks locally and don't need all targets

crates/zk-mobile-bench/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+# OPRF dependencies
+taceo-oprf = { workspace = true }
+
+# Utilities
+rand = { workspace = true }
+rand_chacha = { workspace = true }
+thiserror = { workspace = true }
+
+# UniFFI for mobile bindings
+uniffi = { version = "0.28", features = ["cli"] }
```
paolodamico:
```suggestion
uniffi = { version = "0.31", features = ["cli"] }
```

.github/workflows/compile-gate.yml
```diff
@@ -0,0 +1,31 @@
+name: Compile Gate
```
paolodamico: we should simplify these workflows, this compile gate is doing the same tests that are already being run

dcbuild3r: the compile gate is separate because it won't trigger a browserstack run which can take a while unless the repository first compiles, if other tests are already being run, then can we separate the equivalent of compile gate so I can do if this CI job runs, then trigger this other one, if not don't waste resources?

crates/zk-mobile-bench/docs/benchmark-functions.md
```diff
@@ -0,0 +1,28 @@
+# Benchmark Functions
```
alessandromazza98: you could move this into the already present `README` file

crates/zk-mobile-bench/Cargo.toml
```diff
@@ -0,0 +1,49 @@
+[package]
+name = "zk-mobile-bench"
+version = "0.1.0"
+# Edition 2021 required: `rand`'s `.gen()` conflicts with `gen` keyword reserved in edition 2024.
+edition = "2021"
```
alessandromazza98: we could probably find a workaround for this, but it can be done in a follow up PR

crates/zk-mobile-bench/src/lib.rs
```diff
@@ -0,0 +1,787 @@
...
+        const { RefCell::new(None) };
+}
+
+/// Benchmark: Query Proof (π1) generation only
+///
+/// This benchmarks only the ZK proof generation step, with input and
+/// proving material cached after the first call. Use warmup iterations
+/// to keep setup cost out of measured samples.
+#[benchmark]
+pub fn bench_query_proof_only() {
```
alessandromazza98: what's the difference with the previous benchmark? I don't get it

alessandromazza98 (approved):

PR #416, opened 2026-02-13
fix: don't update oprf key id in RpRegistryWatcher

alessandromazza98:
This PR does the following:

- add `RpRegistry` and `CredentialSchemaIssuerRegistry` to the `make sol-build` command so that when it changes, the abi in the `oprf-node` crate are automatically updated and pull correctly by the Rust code.
- don't update the oprf key id in `RpRegistryWatcher` as it's immutable

---

> [!NOTE]
> **Low Risk**
> Primarily build/ABI plumbing plus a small watcher behavior correction; low risk beyond ensuring the checked-in ABIs match deployed contracts.
> 
> **Overview**
> Ensures `services/oprf-node` uses **checked-in Solidity ABIs** by generating `CredentialSchemaIssuerRegistry` and `RpRegistry` ABI JSONs during `make sol-build` and referencing them directly from the `alloy::sol!` bindings.
> 
> Fixes `RpRegistryWatcher` to **not update** the cached `oprf_key_id` on `RpUpdated` events (treating it as immutable) while still updating the RP signer; includes minor comment/doc formatting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5231b980eccc44c351eed389031d9894d2378b62. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ sol-build:
 	mkdir -p ../crates/authenticator/abi ../crates/issuer/abi ../services/oprf-node/abi && \
 	forge inspect WorldIDRegistry abi --json > ../crates/authenticator/abi/WorldIDRegistryAbi.json && \
 	forge inspect CredentialSchemaIssuerRegistry abi --json > ../crates/issuer/abi/CredentialSchemaIssuerRegistryAbi.json && \
+	forge inspect CredentialSchemaIssuerRegistry abi --json > ../services/oprf-node/abi/CredentialSchemaIssuerRegistryAbi.json && \
 	forge inspect RpRegistry abi --json > ../services/oprf-node/abi/RpRegistryAbi.json
 
 sol-test:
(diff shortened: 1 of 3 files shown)


alessandromazza98:
> idk when the paths to the abi.json files changed, but since it gets updated in the RpRegistry watcher here, should also update it in the CredentialSchemaIssuerRegistry watcher https://github.com/worldcoin/world-id-protocol/blob/ale/fix-rp-registry-watcher/services/oprf-node/src/auth/schema_issuer_registry_watcher.rs#L35 now or in another PR

good point, yes in general I think it's better to have the needed abi for the `oprf-node` inside the `abi` folder in the `oprf-node` crate so that we're always sure that the contracts abis are up to date. I'll update the schema issuer too

PR #433, opened 2026-02-18
feat(indexer): Handle block reorgs

piohei:
This is a first stage. Rollback operation is not yet being called anywhere.
This change requires to nuke db.

---

> [!NOTE]
> **High Risk**
> Introduces breaking DB schema changes and rewires core event persistence/state-application logic, plus new rollback/reorg-detection paths that can delete and replay data if mis-specified. Also adds JSONB-based event storage/decoding, which increases risk of data-shape/compatibility issues during migration and replay.
> 
> **Overview**
> The indexer now persists *full* on-chain World ID Registry events into a new `world_id_registry_events` table (including `block_hash`, `tx_hash`, event type, and JSONB payload), replacing the previous split `world_tree_events`/`world_tree_roots` storage.
> 
> Account writes are updated to track `latest_block_number`/`latest_log_index` on `accounts`, enabling fast identification and deletion of accounts modified after a given event. Event committing is refactored so `EventsCommitter` first performs idempotent event insertion and **fails fast on reorgs** (same `(block_number, log_index)` but different `block_hash`/`tx_hash`), then applies state updates via a new `EventsProcessor`.
> 
> Adds a `RollbackExecutor` that rolls DB state back to a target event by deleting newer events/accounts and replaying per-leaf events up to the rollback point; tree sync/root validation is updated to read from `world_id_registry_events` and handle leaf updates via decoded event details. Tests and helpers are updated accordingly, and legacy world-tree DB accessors are removed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b0370dcb09fd04533090e8ff9c5d9b32951f1bd6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/services/indexer/src/rollback_executor.rs
+++ b/services/indexer/src/rollback_executor.rs
@@ -1,4 +1,7 @@
-use crate::{db::{DB, DBResult, IsolationLevel, PostgresDBTransaction, WorldIdRegistryEventId}, events_processor::EventsProcessor};
+use crate::{
+    db::{DB, DBResult, IsolationLevel, PostgresDBTransaction, WorldIdRegistryEventId},
+    events_processor::EventsProcessor,
+};
 
 pub struct RollbackExecutor<'a> {
     db: &'a DB,
--- a/services/indexer/tests/test_tree_cache.rs
+++ b/services/indexer/tests/test_tree_cache.rs
@@ -9,9 +9,9 @@ use serial_test::serial;
 use std::{fs, path::PathBuf, time::Duration};
 
 use alloy::primitives::{Address, U256, address};
-use world_id_indexer::blockchain::{AuthenticatorRemovedEvent, BlockchainEvent, RegistryEvent};
-use world_id_indexer::config::{
-    Environment, GlobalConfig, HttpConfig, IndexerConfig, RunMode, TreeCacheConfig,
+use world_id_indexer::{
+    blockchain::{AuthenticatorRemovedEvent, BlockchainEvent, RegistryEvent},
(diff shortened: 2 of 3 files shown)


services/indexer/src/rollback_executor.rs
```diff
@@ -0,0 +1,94 @@
...
+                i,
+                event.block_number,
+                event.log_index,
+                event.details
+            );
+        }
+
+        // Apply each event in order
+        for event in events {
+            crate::events_processor::EventsProcessor::process_event(tx, &event).await?;
```
alessandromazza98: nit: you can directly use `EventsProcessor::process_event` for better clarity instead of writing the full path

services/indexer/tests/test_tree_cache.rs
```diff
@@ -489,20 +489,29 @@ async fn test_authenticator_removed_replay() {
...
-    .bind("authentication_removed")
-    .bind(new_commitment_after_removal)
-    .bind((last_block + 1) as i64)
-    .bind(U256::from(1234))
-    .bind(0i64)
-    .execute(&setup.pool)
-    .await
-    .expect("Failed to insert removed event");
+    // Use the proper API instead of raw SQL
+    use world_id_indexer::blockchain::{AuthenticatorRemovedEvent, BlockchainEvent, RegistryEvent};
```
alessandromazza98: nit: the `use` imports can be put in the top of this file

alessandromazza98 (approved):
left some non blocking comments. Approving

PR #436, opened 2026-02-18
feat: explicit proving material loading

Dzejkop:
Removes the implicit embedded groth16 material loading in favour of an explicit setup where the caller must pass it in.

Exposes functions to load & decompress the embedded material - leaving caching up to the consumer of these crates

### DIFF
--- a/crates/core/tests/e2e_authenticator_insert_update_remove.rs
+++ b/crates/core/tests/e2e_authenticator_insert_update_remove.rs
@@ -1,7 +1,6 @@
 #![cfg(feature = "authenticator")]
 
-use std::sync::Arc;
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 
 use alloy::{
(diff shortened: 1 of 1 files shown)


crates/proof/src/proof.rs
```diff
@@ -220,14 +235,14 @@ fn init_circuit_files() -> eyre::Result<CircuitFiles> {
     #[allow(unused_mut)]
     let mut nullifier_zkey = nullifier_zkey.context("OPRFNullifier zkey not found in archive")?;
 
-    // Step 3: ARK decompress zkeys if compress-zkeys is active
+    // Step 3: ARK decompress zkeys if compress-zkeys is active.
```
alessandromazza98: nit: other steps comment don't have the . in the end

alessandromazza98 (approved):
lgtm

Fix ci then good to merge

Issue #428, opened 2026-02-17
Drop non-redis caching


Dzejkop:
Non-redis caching is present only to facilitate local dev & testing - but the services need a database regardless so always end up setting up a testcontainer/running a docker compose project.

We can drop local caching in favour of always using Redis which should simplify the codebase

PR #444, opened 2026-02-19
feat: remove mokka cache from request tracker

kilianglas:
(Partially) addresses https://github.com/worldcoin/world-id-protocol/issues/428. 

Removes the local mokka cache from the request tracker. Tests now require a local Redis instance.


alessandromazza98 (approved):
nice 🚢

PR #447, opened 2026-02-19
fix(authenticator): normalize sparse indexer pubkey slots before key set validation

alessandromazza98:
### Problem

The indexer stores authenticator addresses/pubkeys as sparse slot arrays, where removed authenticators are kept as `null` to preserve stable `pubkey_id` positions.

The previous mapping path compacted these arrays, dropping nulls, which could shift key indices and make gateway/authenticator logic operate on the wrong slot.

### Solution

Preserved sparse slot semantics end-to-end and normalized safely before validation:

- Switched indexer/gateway pubkey payloads to `Vec<Option<U256>>` (and added serde support for optional hex vectors).
- Updated indexer DB mappers to return full slot lists (`None` for removed authenticators) instead of filtering missing entries.
- Added authenticator-side `decode_indexer_pubkeys` normalization:
    - trims trailing empty slots,
    - preserves interior holes to keep slot positions,
    - rejects any used slot beyond `MAX_AUTHENTICATOR_KEYS`.
- Updated authenticator insertion to reuse the first available empty slot instead of always appending.

---

> [!NOTE]
> **Medium Risk**
> Changes affect wire formats and key-slot indexing semantics across services; mismatched deployments or edge cases in slot normalization could break clients or produce incorrect slot selection.
> 
> **Overview**
> Fixes authenticator `pubkey_id` stability by propagating sparse authenticator slot arrays end-to-end instead of compacting away removed entries.
> 
> Indexer API payloads for authenticator pubkeys now return `Vec<Option<U256>>` (serialized via new `hex_u256_opt_vec`), indexer DB mappers stop filtering out `null`s, and inclusion-proof/pubkey routes validate/normalize sparse slots before constructing `AuthenticatorPublicKeySet`.
> 
> On the authenticator side, indexer pubkeys are decoded via `decode_sparse_authenticator_pubkeys` (trim trailing `None`, preserve interior holes, reject out-of-bounds used slots) with a new `InvalidIndexerPubkeySlot` error, and insertion now reuses the first empty slot rather than always appending; tests and stubs are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 226b1194b9a35ee2fc092b69525a80e57e0838b4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9048,7 +9048,6 @@ dependencies = [
  "serde_json",
  "serial_test",
  "sqlx",
- "taceo-ark-babyjubjub",
  "taceo-poseidon2",
  "take_mut",
  "telemetry-batteries 0.2.1",
--- a/services/indexer/Cargo.toml
+++ b/services/indexer/Cargo.toml
@@ -47,7 +47,6 @@ world-id-core = { workspace = true, features = [
   "embed-zkeys",
 ] }
 world-id-primitives = { workspace = true, features = ["openapi"] }
-ark-babyjubjub = { workspace = true }
 
 [dev-dependencies]
 world-id-test-utils = { workspace = true }


crates/primitives/src/authenticator.rs
piohei: I think here it also should be `Option<Vec<Option<EdDSAPublicKey>>>`.

PR #455, opened 2026-02-20
feat(rp): include version byte in rp signature message

Takaros999:
This PR adds a version byte to rp signature message, currently hardcoded to `0x1`


paolodamico (approved):
lgtm! CC @fabian1409 @0xThemis as this affects the auth of OPRF nodes

PR #465, opened 2026-02-21
feat: improve rpc retry behavior

kilianglas:
These changes affect the shared provider module. Currently, it only used in parts of the gateway and indexer services. A later PR will unify usage across the code base. 

Changes:
* Add timeouts for rpc calls
* Add custom exponential backoff retry layer. We extend the retry policy and catch transport errors and some HTTP error codes in addition to the RPC related error the default retry policy catches
* Timeouts and retries are activated by default. Retries can be deactivated by setting max_retries to 0.
* Add integration tests for provider

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.3](https://github.com/worldcoin/world-id-protocol/compare/world-id-primitives-v0.4.2...world-id-primitives-v0.4.3) - 2026-02-23
+
+### Added
+
+- *(rp)* include version byte in rp signature message
+
+### Other
+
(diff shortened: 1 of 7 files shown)


services/common/src/provider.rs
```diff
@@ -219,21 +281,56 @@ impl ProviderArgs {
...
+                .layer(retry_layer)
+                .transport(transport, false)
         } else {
             let transport = ServiceBuilder::new()
                 .layer(fallback_layer)
                 .service(transports);
-
-            RpcClient::builder().transport(transport, false)
+            RpcClient::builder()
+                .layer(retry_layer)
```
0xOsiris: Nit: It'd be better to `option_layer` imo on the `None` variant of `Option<RetryConfig>`

right now it looks like there's no path to not configure the retry/timeout layer (which may be fine for our use)

0xOsiris (approved):
Just one suggestion. Nice work!

services/common/src/provider.rs
kilianglas: The idea is to enable retries per default. Will make it explicit to disable, though

PR #494, opened 2026-02-25
feat: gateway orphan sweeper

kilianglas:
## Summary

Currently, gateway requests can become orphaned when a replica crashes or restarts between accepting a request and confirming its on-chain transaction, because requests states are only updated by the replica that handles them. Without cleanup, these requests remain stuck in `Queued`, `Batching`, or `Submitted` state indefinitely, leaving clients with no resolution.

This PR introduces an **orphan sweeper** -- a background task that periodically scans all pending requests in Redis and resolves those that have gone stale.

For simplicity, the orphan sweeper runs on every replica and every operation that touches Redis is idempotent.

### Sweeper design

We now maitain a `gateway:pending_requests` in Redis. Requests get added atomically when they are added in Queued state. Likewise, they are removed when the status is updated to Finalized or Failed.

The sweeper runs on a configurable interval (default: 30s) and operates in two phases:

1. **Snapshot**: Loads all request IDs from the `gateway:pending_requests` Redis set and bulk-fetches their records via `MGET`.
2. **Resolution**: For each pending request:
   - **Dangling set entries** (no record in Redis): removed from the pending set. This should never happen in practice.
   - **Already terminal** (`Finalized`/`Failed`): removed from the pending set (stale cleanup). Again, this should never happen in practice since pending set entries are removed atomically when the corresponding request status is updated to finalized or failed.
   - **`Queued`/`Batching`** older than `stale_queued_threshold_secs` (default: 60s): marked as `Failed` ("request orphaned").
   - **`Submitted`** with a tx hash: groups by tx hash, performs deduplicated receipt lookups. If a receipt exists, finalizes based on `receipt.status()`. If no receipt and older than `stale_submitted_threshold_secs` (default: 600s): marked as `Failed` ("transaction not confirmed").

### Changes

- **`orphan_sweeper.rs`** (new): Implements `run_orphan_sweeper` (background loop) and `sweep_once` (single pass, exposed for testing).
- **`request_tracker.rs`**:
  - `new_request_with_id` now uses an atomic Lua script for `SET NX EX` + `SADD` to the pending set, preventing partial writes if the process crashes between the two operations.
  - `set_status_on_redis` atomically removes from the pending set when transitioning to a terminal state (`Finalized`/`Failed`) via Lua script.
  - Added `finalize_from_receipt`, `snapshot_batch`, `get_pending_requests`, and `remove_from_pending_set` helpers.
- **`config.rs`**: Added `OrphanSweeperConfig` with `clap::Args` derive and `Default` impl, embedded in `GatewayConfig` via `#[command(flatten)]`.
- **`create_batcher.rs` / `ops_batcher.rs`**: Refactored receipt-checking spawned tasks to use the shared `finalize_from_receipt` method.
- **`routes.rs`**: Wires up the sweeper as a background task in `build_app`.
- **Tests**: 12 integration tests covering the full sweeper behavior (stale/fresh for each state, dangling entries, terminal cleanup, real on-chain receipt resolution) plus pending set lifecycle and `updated_at` timestamp tests.

### Breaking Changes

We add the updated_at field to the request records in Redis. Hence, when we update on stage, the status endpoint will not work for request records currently present in Redis. This should not be a big deal.

### Future Work

This PR adds a new nested config struct. A future PR will refactor the entire config (e.g., batcher config) to stay consistent with this pattern. This also get's rid of the surpressed clippy warning in run_app.

### DIFF
--- a/services/gateway/src/request_tracker.rs
+++ b/services/gateway/src/request_tracker.rs
@@ -257,9 +257,7 @@ impl RequestTracker {
                 let record = maybe_json.and_then(|json_str| {
                     serde_json::from_str::<RequestRecord>(&json_str)
                         .map_err(|e| {
-                            tracing::error!(
-                                "Failed to deserialize request {id} from Redis: {e}"
-                            );
+                            tracing::error!("Failed to deserialize request {id} from Redis: {e}");
                         })
                         .ok()
                 });


services/gateway/src/request_tracker.rs
```diff
@@ -140,22 +195,99 @@ impl RequestTracker {
...
+    /// Returns all request IDs currently in the pending set.
+    pub async fn get_pending_requests(&self) -> Vec<String> {
+        let mut manager = self.redis_manager.clone();
+        let result: Result<std::collections::HashSet<String>, redis::RedisError> =
+            manager.smembers(PENDING_SET_KEY).await;
+
+        match result {
+            Ok(ids) => ids.into_iter().collect(),
+            Err(e) => {
+                tracing::error!("Failed to get pending requests from Redis: {e}");
```
alessandromazza98: is this a recoverable error? otherwise it could make sense to return an error (??)

services/gateway/tests/test_orphan_sweeper.rs
```diff
@@ -0,0 +1,575 @@
...
+            &id,
+            GatewayRequestState::Submitted {
+                tx_hash: "0xdef".to_string(),
+            },
+        )
+        .await;
+
+    let record = read_raw_record(&mut redis, &id).await.unwrap();
+    let submitted_at = record["updated_at"].as_u64().unwrap();
+    assert!(submitted_at >= created_at);
```
alessandromazza98: why not only > ?

alessandromazza98 (approved):
left several comments. Approving to not block because the design makes sense to me. Feel free to address the comments I left

services/gateway/src/orphan_sweeper.rs
```diff
@@ -0,0 +1,137 @@
...
+                    .or_default()
+                    .push((id.clone(), record.updated_at));
+            }
+        }
+    }
+
+    // Phase 2: deduplicated receipt lookups
+    for (tx_hash, group) in &submitted_groups {
+        let Ok(hash) = tx_hash.parse::<B256>() else {
+            tracing::error!(tx_hash = %tx_hash, "sweeper: invalid tx_hash, skipping group");
```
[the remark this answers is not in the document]
kilianglas: I think you are right, the corret behavior is to evict to prevent the same error from happening over and over

PR #496, opened 2026-02-25
fix(gateway): Redis-backed nonce manager for multi-replica safety

### LINKED ISSUE #495: Gateway: `nonce too low` when running multiple replicas with shared signer
## Title
Gateway: `nonce too low` when running multiple replicas with shared signer

## Problem
In stage, `world-id-gateway` sometimes fails account registration with:

`server returned an error response: error code -32000: nonce too low: next nonce 82, tx nonce 81`

Datadog logs showing this error:
[Stage gateway error logs (`nonce too low`)](https://app.datadoghq.com/logs?query=service%3Aworld-id-gateway%20env%3Astage%20cluster_name%3Acrypto-v2-stage-us-east-1%20status%3Aerror&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZyREzI7FHQymgAAABhBWnlSRXoyekFBQmhBM0c5VmNDbjh3QUYAAAAkZjE5YzkxMTMtN2YxZi00OTc5LWExYTYtNDZkZGZjZTI5NTNhAAIOSg&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1771702856035&to_ts=1771962056035&live=true)

## Root cause
Gateway uses process-local nonce management (`CachedNonceManager`).
With 2+ replicas sharing the same signer key, nonce allocation is not coordinated across replicas, so collisions occur.

Relevant code:
- `services/common/src/provider.rs` (uses `CachedNonceManager`)
- `services/gateway/src/create_batcher.rs` (`createManyAccounts(...).send().await`)
- `services/gateway/src/ops_batcher.rs` (also submits txs, consumes nonces)

## Expected
No nonce collisions in multi-replica deployment.

## Actual
Intermittent `nonce too low` tx send failures.

## Proposed fix
- Short term: run one tx-submitting replica per signer (or unique signer per replica).
- Long term: add distributed nonce coordination (or a single tx-submitter component).

## Acceptance criteria
- [ ] No `nonce too low` failures when running multiple replicas.

Dzejkop:
## Summary

Fixes #495 — **`nonce too low` tx send failures** when multiple gateway replicas share one signer.

### Root Cause

`CachedNonceManager` (alloy built-in) maintains an in-process nonce counter. When 2+ replicas share the same signer key, each independently tracks nonces → collisions.

### Changes

| File | What |
|------|------|
| `services/common/src/provider.rs` | Add `http_with_nonce_manager<M>()` to `ProviderArgs` — callers can now inject any `NonceManager` implementation |
| `services/common/src/lib.rs` | Re-export `NonceManager` trait for downstream use |
| `services/gateway/src/nonce.rs` | **New** — `RedisNonceManager` implementing alloy's `NonceManager` trait |
| `services/gateway/src/lib.rs` | Wire `RedisNonceManager` into provider construction |
| `services/gateway/Cargo.toml` | Add `async-trait` dep (needed for `NonceManager` impl) |

### How RedisNonceManager works

A single Lua script runs atomically in Redis:

1. `SET key (on_chain_nonce - 1) NX` — seed from on-chain pending count on first access only
2. `INCR key` — return next unique nonce
3. **Safety floor** — if Redis value < on-chain nonce (stale data), bump up

This guarantees **every replica gets a unique, monotonically increasing nonce** regardless of how many processes call it concurrently.

### No new configuration

Reuses the existing `REDIS_URL` environment variable. Zero config change required for rollout.

### Tests

- **Sequential**: verifies nonces increment correctly from on-chain seed
- **Safety floor**: verifies recovery when on-chain nonce jumps ahead of Redis
- **Concurrent** (50 tasks): verifies all nonces are unique and sequential under contention

### Rollout Notes

- **Safe to deploy with existing config** — no new env vars needed.
- Rolling restart is fine — Redis nonce state persists across replica restarts.
- If Redis is flushed, the nonce re-seeds from the on-chain pending count automatically.
- Backward compatible: `ProviderArgs::http()` still works with the default `CachedNonceManager` for services that don't need distributed coordination.

Co-authored-by: otto@toolsforhumanity.com

### DIFF
--- a/services/gateway/src/error.rs
+++ b/services/gateway/src/error.rs
@@ -60,6 +60,12 @@ pub enum GatewayError {
     },
     #[error("config error: {0}")]
     Config(String),
+    #[error("redis nonce manager error: {source}")]
+    RedisNonceManager {
+        #[source]
+        source: redis::RedisError,
+        backtrace: String,
+    },
 }
 
 impl From<ProviderError> for GatewayError {
--- a/services/gateway/src/lib.rs
+++ b/services/gateway/src/lib.rs
@@ -115,11 +115,18 @@ pub async fn run() -> GatewayResult<()> {
     // Use Redis-backed nonce manager so multiple replicas sharing the same
     // signer key never collide on nonces.  The existing REDIS_URL config
     // value is reused — no new configuration required.
-    let redis_client = redis::Client::open(cfg.redis_url.as_str())
-        .map_err(|e| GatewayError::Config(format!("invalid REDIS_URL for nonce manager: {e}")))?;
-    let redis_conn = ConnectionManager::new(redis_client).await.map_err(|e| {
-        GatewayError::Config(format!("Redis connection for nonce manager failed: {e}"))
+    let redis_client = redis::Client::open(cfg.redis_url.as_str()).map_err(|source| {
+        GatewayError::RedisNonceManager {
+            source,
+            backtrace: Backtrace::capture().to_string(),
+        }
     })?;
+    let redis_conn = ConnectionManager::new(redis_client)
+        .await
+        .map_err(|source| GatewayError::RedisNonceManager {
+            source,
+            backtrace: Backtrace::capture().to_string(),
+        })?;
     let nonce_mgr = RedisNonceManager::new(redis_conn);
     tracing::info!("Redis-backed nonce manager initialised");
 


alessandromazza98:
> What is the advantage of this over having a signer per pod?

I'd say main advantage is you only need to handle one address to refill and monitor. I'm not against having different signers per pod, but since the average we expect (after initial migration) is ~3000 requests per day, it probably makes sense to just have one signer to handle it.

PR #557, opened 2026-03-13
draft: action prefix  

Takaros999:
Adds an example test to demonstrate how setting the MSB to `0x01` would still fit into a fit element.

https://www.notion.so/worldcoin/Temp-OPRF-Nodes-for-Session-ID-Seed-3188614bdf8c80d48ff7cd2b95a17186?source=copy_link

image

### DIFF
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -529,4 +529,66 @@ mod tests {
         let back: U256 = fe.into();
         assert_eq!(original, back);
     }
+
+    /// Illustrates the session action prefix scheme for the spec.
+    ///
+    /// Hashed actions use `keccak256(string) >> 8` which always produces a top byte of 0x00.
+    /// Session actions set the top byte to 0x01, which is still a valid field element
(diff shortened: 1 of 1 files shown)


crates/primitives/src/lib.rs
```diff
@@ -529,4 +529,66 @@ mod tests {
...
+    fn test_session_action_prefix_scheme() {
+        // --- Hashed actions always have MSB = 0x00 ---
+        let action_strings = [
+            "verify-humanity",
+            "login",
+            "vote-proposal-42",
+            "some-really-long-action-string-that-could-be-anything",
+        ];
+        for action_str in action_strings {
+            let fe = FieldElement::from_arbitrary_raw_bytes(action_str.as_bytes());
```
paolodamico: we should also have some examples using ABI-encoding like here, https://github.com/worldcoin/world-id-utility-contracts/blob/89483e9bbe64da9e29b1f9254b28cec81be73133/src/address-book/AddressBook.sol#L260

PR #608, opened 2026-03-25
refactor(oprf-node): misc stuff for oprf-nodes (v1.1)

0xThemis:
- **chore(oprf-node): updated a doc line**
- **test(oprf-node): added some more tests for auth modules**
- **feat(oprf-node): check also expiration time on RP signature**
- **refactor(oprf-node)!: added TTL/TTI config for caches**
- **refactor(oprf-node): use one provider over multiple**

### DIFF
--- a/services/oprf-node/src/auth/rp_module.rs
+++ b/services/oprf-node/src/auth/rp_module.rs
@@ -874,10 +874,7 @@ mod tests {
             auth_error.code(),
             primitives::oprf::error_codes::INVALID_ACTION_NULLIFIER
         );
-        assert_eq!(
-            auth_error.message(),
-            "invalid action - MSB must be 0x00 for nullifier"
-        );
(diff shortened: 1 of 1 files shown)


services/oprf-node/src/config.rs
```diff
@@ -63,20 +59,62 @@ pub struct WorldOprfNodeConfig {
...
+    #[serde(default = "WatcherCacheConfig::default_time_to_idle")]
+    pub time_to_idle: Duration,
+}
 
-    /// Default maximum RP registry store size
-    const fn default_max_rp_registry_store_size() -> u64 {
+impl WatcherCacheConfig {
+    /// Default maximum size of the cache
+    const fn default_max_cache_size() -> u64 {
         1000
```
paolodamico: this seems small for a production setup

0xThemis: Will use 10k for production for both caches 👍

PR #613, opened 2026-03-25
fix(contracts): root valid ttl race condition

0xThemis:
- **feat: introduce V2 verifier enforcing action check**
- **feat(contracts): roots TTL start now when replaced over inserted**


contracts/src/core/WorldIDRegistryV2.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: MIT
```
paolodamico: let's follow the existing pattern and name this file `WorldIDRegistryV2Unreleased.sol` until it's released

PR #643, opened 2026-04-06
feat: WIP-104 - Proving and Admin Authenticators with Different Permission Sets

paolodamico:
Changes and rationale in the WIP-104 spec.

---

> [!NOTE]
> **High Risk**
> High risk because it changes on-chain authenticator management and authorization semantics (including disabling `updateAuthenticator` and introducing a new bitmap encoding with a lower hard limit), which can affect account recoverability and key-management flows.
> 
> **Overview**
> Implements WIP-104 in `WorldIDRegistryV2Unreleased` by introducing **two authenticator classes**: Admin authenticators (with on-chain management address) and Proving authenticators (no management address, encoded via a class bit in the existing pubkey bitmap). `insertAuthenticator` now allows `address(0)` to create a proving authenticator, and `removeAuthenticator` branches on the class bit to validate/clean up mappings appropriately while enforcing a new *manageability invariant* (cannot remove the last admin if proving authenticators remain).
> 
> V2 also **disables** `updateAuthenticator` by reverting with `MethodUnsupported`, and overrides `setMaxAuthenticators` to enforce a new hard cap of **48** (bitmap split between occupancy and class flags). Adds `IWorldIDRegistryV2` (new errors), comprehensive Forge tests for the WIP-104 behavior, updates the race-condition test to reference the new `UnknownRoot` error, and makes small overrideability/doc tweaks (`virtual` on select V1 methods; README guidance; minor Rust doc wording).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19cbe6894960d52005e86c4e40d267739ad4ffa9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


paolodamico:
@codex review

paolodamico:
cursor review

contracts/src/core/WorldIDRegistryV2Unreleased.sol
```diff
@@ -58,4 +85,229 @@ contract WorldIDRegistryV2 is WorldIDRegistry {
...
+        uint256 newOffchainSignerCommitment,
+        bytes memory signature,
+        uint256 nonce
+    ) external virtual override(IWorldIDRegistry, WorldIDRegistry) onlyProxy onlyInitialized {
+        if (newAuthenticatorAddress != address(0)) {
+            _validateNewAuthenticatorAddress(newAuthenticatorAddress);
+        }
+
+        if (pubkeyId >= _maxAuthenticators) {
+            revert PubkeyIdOutOfBounds();
```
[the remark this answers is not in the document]
paolodamico: This gets manually reviewed on deployment. Current value is `7`.

docs/WIPs/wip-104.md
```diff
@@ -0,0 +1,107 @@
...
+
+Prior to the introduction of this spec, all registered authenticators MUST be considered Admin Authenticators.
+
+Going forward, "management operations" refers to any operation which results in a state change to a user's World ID in the `WorldIDRegistry`. Examples: inserting an authenticator, removing an authenticator, initiating a Recovery Agent update.
+
+### Bitmap Encoding
+
+The `WorldIDRegistry` stores a 96-bit pubkey bitmap per account in `_leafIndexToRecoveryAddressPacked`. This spec splits the bitmap into two halves:
+- **Bits [0-47]**: Occupancy. Bit `n` is set if pubkeyId `n` is in use.
+- **Bits [48-95]**: Authenticator Dlass. Bit `48 + n` is set if `pubkeyId` `n` is a Proving Authenticator.
```
kilianglas:
```suggestion
- **Bits [48-95]**: Authenticator Class. Bit `48 + n` is set if `pubkeyId` `n` is a Proving Authenticator.
```

kilianglas:
What if we completely decouple admin keys from proving keys? E.g., I might want to keep a backup admin key around on a cold wallet if I loose access to all of my other authenticators, but I would never use this for proofs.

alessandromazza98:
> What if we completely decouple admin keys from proving keys? E.g., I might want to keep a backup admin key around on a cold wallet if I loose access to all of my other authenticators, but I would never use this for proofs.

what is the purpose though? an authenticator that has management over the `WorldIdRegistry` is by construction an admin authenticator because it can adds new authenticators, it can remove them and so on.

By the way, I generally like this WIP.

paolodamico:
@codex review

docs/WIPs/wip-104.md
```diff
@@ -0,0 +1,128 @@
...
+
+The operation MUST be authorized by an existing Admin Authenticator (no changes).
+
+### `updateAuthenticator`
+
+The `updateAuthenticator` function is deprecated in this version. Calls MUST revert with `MethodUnsupported`. Key rotation can be accomplished with a `removeAuthenticator` followed by `insertAuthenticator`.
+
+### Account Creation
+
+Account creation (`_registerAccount`) is unaffected. `_registerAccount` rejects `address(0)`, so every new account starts with at least one Admin Authenticator. As a consequence, an account can never be created with only Proving Authenticators, even though that is a legal end-state reachable via `insertAuthenticator` + `removeAuthenticator`.
```
0xOsiris: > even though that is a legal end-state reachable via `insertAuthenticator` + `removeAuthenticator`

This is by design? Should this be the case?

docs/WIPs/wip-104.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Manageability Invariant
+
+An Admin Authenticator MUST NOT be removed if it is the only Admin Authenticator remaining on the account. Such a call MUST revert with `UnmanageableNotAllowed`. This guarantees that an account with one or more Proving Authenticators always retains at least one Admin Authenticator capable of managing it.
```
0xOsiris: I believe this directly contradicts the statement above?

docs/WIPs/wip-104.md
```diff
@@ -0,0 +1,128 @@
...
+
+A general note applies to both options: practical accounts are expected to hold a small number of authenticators (the V1 default for `_maxAuthenticators` is 7 and has never been raised), so the 48-slot cap imposed by the chosen encoding is not a binding constraint. Most importantly, there is no concrete use case that warrants having standalone Management Keys. In the **future** there is a plausible evolution where Management Keys exist on Authenticator-like primitives that offer alternative mechanisms for users to hold their World ID Keys (examples: passkeys, zkLogin alternatives, etc). These constructs could provide recovery alternatives to users. This direction is not pursued in this spec because no such admin type exists. At this time there are off-protocol mechanisms (e.g. authenticator-managed backups) already address most of the same user pain points.
```
0xOsiris: > In the **future** there is a plausible evolution where Management Keys exist on Authenticator-like primitives that offer alternative mechanisms for users

The wording here makes this sound like the utility is still unknown, but in my opinion this is clearly the correct direction for abstraction of the authentication primitives. Might be better to word it as out of scope, but planned.

contracts/src/core/WorldIDRegistryV2Unreleased.sol
```diff
@@ -58,4 +85,229 @@ contract WorldIDRegistryV2 is WorldIDRegistry {
...
+    function insertAuthenticator(
+        uint64 leafIndex,
+        address newAuthenticatorAddress,
+        uint32 pubkeyId,
+        uint256 newAuthenticatorPubkey,
+        uint256 oldOffchainSignerCommitment,
+        uint256 newOffchainSignerCommitment,
+        bytes memory signature,
+        uint256 nonce
+    ) external virtual override(IWorldIDRegistry, WorldIDRegistry) onlyProxy onlyInitialized {
```
0xOsiris: I still don't understand why we use `onlyInitialized`. The proxy is already initialized. This could never fail; It just adds an SLOAD to every function call

contracts/src/core/WorldIDRegistryV2Unreleased.sol
```diff
@@ -58,4 +85,229 @@ contract WorldIDRegistryV2 is WorldIDRegistry {
...
+    function removeAuthenticator(
+        uint64 leafIndex,
+        address authenticatorAddress,
+        uint32 pubkeyId,
+        uint256 authenticatorPubkey,
+        uint256 oldOffchainSignerCommitment,
+        uint256 newOffchainSignerCommitment,
+        bytes memory signature,
+        uint256 nonce
+    ) external virtual override(IWorldIDRegistry, WorldIDRegistry) onlyProxy onlyInitialized {
```
0xOsiris: Maybe I'm overlooking it, but shouldn't we be constraining the type value of `pubkeyId` to `<= 48` or `_maxAuthenticators`?

0xOsiris (dismissed):
Left some suggestions, and general notes

contracts/src/core/WorldIDRegistryV2Unreleased.sol
kilianglas: Same for me

docs/WIPs/wip-104.md
kilianglas: This is statement is also confusing to me. I'd probably remove `even though that is a legal end-state reachable via `insertAuthenticator` + `removeAuthenticator`.`

docs/WIPs/wip-104.md
```diff
@@ -0,0 +1,128 @@
...
+The `updateAuthenticator` function is deprecated in this version. Calls MUST revert with `MethodUnsupported`. Key rotation can be accomplished with a `removeAuthenticator` followed by `insertAuthenticator`.
+
+### Account Creation
+
+Account creation (`_registerAccount`) is unaffected. `_registerAccount` rejects `address(0)`, so every new account starts with at least one Admin Authenticator. As a consequence, an account can never be created with only Proving Authenticators, even though that is a legal end-state reachable via `insertAuthenticator` + `removeAuthenticator`.
+
+### Manageability Invariant
+
+An Admin Authenticator MUST NOT be removed if it is the only Admin Authenticator remaining on the account. Such a call MUST revert with `UnmanageableNotAllowed`. This guarantees that an account with one or more Proving Authenticators always retains at least one Admin Authenticator capable of managing it.
+
```
kilianglas, line 88: Shouldn't this be `and account with ZERO or more Proving Authenticators ...`

contracts/src/core/WorldIDRegistryV2Unreleased.sol
Kemperino: I think it is implicitly checked.

since first we check if its provingAuthenticators
`        bool isProvingAuthenticator = (bitmap & (1 << (_AUTHENTICATOR_CLASS_OFFSET + pubkeyId))) != 0;
`
for admin authenticators we fetch packedToRemove and then compare the pubkeyId.

I think constraining **<** 48 for clarification makes sense though as well.

docs/WIPs/wip-104.md
```diff
@@ -0,0 +1,128 @@
...
+2. **Admin Authenticator**: Holds **both** an **Signing Key** and a **Management Key**. An Admin Authenticator is capable of all operations for a user's World ID, including management operations and proof generation.
+
+Prior to the introduction of this spec, all registered authenticators MUST be considered Admin Authenticators.
+
+### Bitmap Encoding
+
+The `WorldIDRegistry` stores a 96-bit pubkey bitmap per account in `_leafIndexToRecoveryAddressPacked`. This spec splits the bitmap into two halves:
+- **Bits [0-47]**: Occupancy. Bit `n` is set if pubkeyId `n` is in use.
+- **Bits [48-95]**: Authenticator Class. Bit `48 + n` is set if `pubkeyId` `n` is a Proving Authenticator.
+
```
Kemperino, line 50: We could in theory also define the type of Authenticator in a 2 -tuple, where
(0,0)=empty
(1,0)= admin
(0,1)=proving
(1,1)= another type of management key

Kemperino (dismissed):

docs/WIPs/wip-104.md
paolodamico: actually this state cannot be reached, you can't remove the last admin authenticator. updated to clarify

docs/WIPs/wip-104.md
paolodamico: extremely fair, updated

contracts/src/core/WorldIDRegistryV2Unreleased.sol
paolodamico: actually I don't think this is a good idea because if the maxAuthenticators are reduced it may leave some authenticators that can't be removed. we do check this when inserting though

docs/WIPs/wip-104.md
paolodamico: yes 100%, made it clearer

docs/WIPs/wip-104.md
paolodamico: totally, in the end that's roughly the layout we have, we're just doing a different bit allocation, and (0,1) is the undefined one.

contracts/src/core/WorldIDRegistryV2Unreleased.sol
0xOsiris: Yes, this is why I suggested 48. As this is the least upper bound as I understand it regardless of `_maxAuthenticators`. Although I'm not certain if it matters other than being explicit on type constraints which is generally best practice

0xOsiris (approved):

PR #648, opened 2026-04-10
feat: ownership proof generation

paolodamico:
Exposes `generate_ownership_proof` which allows for proof generation of the "Ownership Proof" using ProveKit as the ZK backend.

Particularly, proving and verifying keys are generated at build time 🙌, no more downloading artifacts at build time.

---

> [!NOTE]
> **Medium Risk**
> Adds a new ProveKit/Noir proof-generation path with build-time `nargo` compilation and substantial new dependencies, which can affect build reproducibility and CI stability. Also updates CI toolchains (MSRV/test matrix) and dependency lockfile, so failures are more likely than a pure refactor.
> 
> **Overview**
> Adds *experimental* support for generating WIP-103 “ownership proofs” using ProveKit + a Noir circuit. This introduces a new `OwnershipProofCircuitInput` (in primitives) and a new `generate_ownership_proof` implementation in `world-id-proof` that embeds a generated ProveKit proving package and converts Rust inputs into a Noir witness.
> 
> Build/CI is updated to support this flow: `world-id-proof` gains a `provekit` feature and ProveKit deps, `build.rs` now runs `nargo compile` to generate `ownership_proof.pkp` at build time (skipped on docs.rs), and GitHub Actions installs the pinned Noir toolchain (`v1.0.0-beta.11`) across Rust/Noir workflows; Rust CI also bumps MSRV to `1.93.0` and runs tests on both `stable` and `nightly`.
> 
> Also includes small cleanup/fixes: `ProofError` is moved/re-exported at the crate root (updating imports/docs), `cargo-deny` advisory ignores are adjusted for new ProveKit transitive deps, and relay backfill progress computation is made overflow/zero-safe via `checked_div`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63a57a5a714d94cf1da19c36e4db34c44462d2fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #661, opened 2026-04-15
fix: drop version from unpublished world-id-test-utils workspace dep

### LINKED ISSUE #660: world-id-proof can't publish to crates.io: unpublished dev-dep on world-id-test-utils
## Problem

`world-id-proof` has a dev-dependency on `world-id-test-utils` (added in #648), which is not published to crates.io. This causes `cargo publish` to fail:

```
error: failed to prepare local package for uploading

Caused by:
  no matching package named `world-id-test-utils` found
  location searched: crates.io index
  required by package `world-id-proof v0.9.0`
```

This blocks the v0.9.0 release (and all future releases). Currently only `world-id-primitives` is published at 0.9.0; the other four crates are stuck at 0.8.2.

## Context

- The v0.9.0 release PR (#639) was merged, but only `world-id-primitives` got published due to a release-plz bug (fixed in #659)
- After the fix, release-plz correctly skips the already-published primitives but fails on `world-id-proof` due to this dep issue

## Options

1. Publish `world-id-test-utils` to crates.io
2. Exclude `world-id-test-utils` from the published package (e.g., make it a path-only dep not resolved from the registry)

## Affected file

`crates/proof/Cargo.toml:70`

```toml
[dev-dependencies]
world-id-test-utils = { workspace = true }
```

kilianglas:
## Summary

- Removes `version = "0.1.0"` from the `world-id-test-utils` workspace dependency definition
- This unblocks `cargo publish` for `world-id-proof` (and all other published crates)

## Problem

PR #648 added a dev-dependency on `world-id-test-utils` to `world-id-proof`. Since the workspace dep included `version = "0.1.0"`, `cargo publish` tries to resolve it from crates.io — but `world-id-test-utils` has `publish = false` and will never exist there.

This wasn't an issue before v0.9.0 because `world-id-proof` had no dev-dependencies prior to #648.

## Fix

Dev-dependencies with only a `path` (no `version`) are stripped from the published manifest by cargo. Removing the version from the workspace definition is sufficient — the path alone is all that's needed for local resolution.

Closes #660

## Test plan

- [x] `cargo check -p world-id-proof` passes
- [x] `cargo publish --dry-run -p world-id-proof` no longer fails on the missing `world-id-test-utils` dep


alessandromazza98 (approved):

Dzejkop (approved):

alessandromazza98:
there is `deny` failing; it suggests to update `cargo update -p rustls-webpki`

PR #691, opened 2026-04-24
build: remove oprf-key-gen dep and use testcontainers

0xThemis:


crates/test-utils/src/stubs.rs
```diff
@@ -323,181 +320,200 @@ pub async fn spawn_oprf_nodes(
     .into()
 }
 
+const OPRF_KEY_GEN_IMAGE: &str = "ghcr.io/taceolabs/oprf-service/oprf-key-gen";
+const OPRF_KEY_GEN_TAG: &str = "v1.1.0-rc.3";
```
dkales: Is this not an old one?

crates/test-utils/src/stubs.rs
```diff
@@ -323,181 +320,200 @@ pub async fn spawn_oprf_nodes(
...
+            "TACEO_OPRF_KEY_GEN__SERVICE__ZKEY_PATH",
+            "/app/circom/OPRFKeyGen.25.arks.zkey",
+        )
+        .with_env_var(
+            "TACEO_OPRF_KEY_GEN__SERVICE__WITNESS_GRAPH_PATH",
+            "/app/circom/OPRFKeyGenGraph.25.bin",
+        )
+        .with_env_var(
+            "TACEO_OPRF_KEY_GEN__SERVICE__CONFIRMATIONS_FOR_TRANSACTION",
+            "1",
```
dkales, line 401: We have those in the docker image anyway, no?

crates/test-utils/src/stubs.rs
0xThemis: Yes was the current one in tests - will update in this PR

crates/test-utils/src/stubs.rs
0xThemis: if you mean the artifacts (zkey/path) you are right - also removed the mounting from the docker-compose. Was there for a long time I think and was copied now to this pass

dkales (approved):

alessandromazza98 (approved):

PR #695, opened 2026-04-27
fix(oprf-node): propagate cancellation to cache maintenance tasks

alessandromazza98:
### Problem

Each watcher (`MerkleWatcher`, `RpRegistryWatcher`, `SchemaIssuerRegistryWatcher`) and `NonceHistory` in the OPRF node spawn a background cache maintenance task that runs `run_pending_tasks` and updates metrics on a fixed interval. Unlike the corresponding subscribe tasks, these maintenance tasks:

- did not check the shared `CancellationToken`, so they kept running until the tokio runtime was dropped at process exit.
- had their `JoinHandles` dropped immediately after spawning, so panics or unexpected exits were silently swallowed instead of surfacing during graceful shutdown.
- in the case of `NonceHistory::init`, did not even receive a `CancellationToken`, so there was no way to wire one in.
This meant the service's shutdown path `WorldOprfNodeTasks::join` only observed the subscribe tasks, leaving the maintenance tasks untracked and unobservable.

### Solution

Introduced a small `BackgroundTask` alias and a `WatcherBackgroundTasks { subscribe, maintenance }` struct to keep watcher init signatures readable. Extracted each maintenance loop into a top-level `maintenance_task` function that:

- installs a `cancellation_token.drop_guard()` so a panic triggers global shutdown, mirroring the subscribe tasks.
- uses `tokio::select!` between `interval.tick()` and `cancellation_token.cancelled()` to exit cleanly on shutdown.
- added a `cache_maintenance_tasks: Vec<BackgroundTask>` field to `WorldOprfNodeTasks`, collected every maintenance handle in `start`, and joined them via `futures::future::join_all` in `WorldOprfNodeTasks::join` so panics and errors are now propagated rather than silently lost.

---

> [!NOTE]
> **Medium Risk**
> Changes how long-running watcher/nonce cache maintenance tasks are spawned, cancelled, and awaited; mistakes here could cause hangs or missed shutdown/error propagation. No changes to auth/verification logic or on-chain semantics.
> 
> **Overview**
> Improves graceful shutdown observability by **tracking and cancelling cache maintenance loops** for `MerkleWatcher`, `RpRegistryWatcher`, `SchemaIssuerRegistryWatcher`, and `NonceHistory`.
> 
> Watcher `init` methods now return `WatcherBackgroundTasks { subscribe, maintenance }`, maintenance loops are extracted into cancellable tasks (using `tokio::select!` on `CancellationToken`) and guarded so panics/unexpected exits trigger shutdown. `WorldOprfNodeTasks` collects all maintenance `JoinHandle`s and joins them in `WorldOprfNodeTasks::join`, ensuring errors/panics surface instead of being silently dropped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09dec8d55d0c738f0fcb0afe254be1636898623d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/services/oprf-node/src/auth.rs
+++ b/services/oprf-node/src/auth.rs
@@ -24,6 +24,20 @@ pub(crate) mod rp_module;
 pub(crate) mod rp_registry_watcher;
 pub(crate) mod schema_issuer_registry_watcher;
 
+/// A handle to a background task that returns `eyre::Result<()>`.
+pub(crate) type BackgroundTask = tokio::task::JoinHandle<eyre::Result<()>>;
+
+/// Handles to the long-running background tasks spawned by a watcher's `init`
+/// method: a subscribe task that processes contract events and a cache
+/// maintenance task that runs periodic upkeep on the watcher's cache. Both
+/// tasks observe the watcher's [`tokio_util::sync::CancellationToken`] and
+/// should be awaited during graceful shutdown so that panics or unexpected
+/// errors are surfaced rather than silently dropped.
+pub(crate) struct WatcherBackgroundTasks {
+    pub(crate) subscribe: BackgroundTask,
+    pub(crate) maintenance: BackgroundTask,
+}
+
 pub(crate) fn verify_query_proof(
     vk: &PreparedVerifyingKey<Bn254>,
     proof: &ark_groth16::Proof<Bn254>,
@@ -275,9 +289,10 @@ mod tests {
             )
             .await?;
 
-            let nonce_history = NonceHistory::init(
+            let (nonce_history, _) = NonceHistory::init(
                 current_time_stamp_max_difference * 2,
                 cache_maintenance_interval,
+                cancellation_token.clone(),
             );
 
             Ok(Self {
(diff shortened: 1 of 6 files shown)


0xThemis:
after some discussions after the rust audit findings we agreed to remove the current implementation for the caches and move to a completly lazy cache without on-chain event streaming:

https://github.com/worldcoin/world-id-protocol/pull/658/changes/076307958ea6fb3d3a2590fa86dc2f0b9ef3749f

so this is not needed anymore - we also removed the maintaince tasks completly removing the necessacity for this PR. Would close this if that is fine with you

alessandromazza98:
> after some discussions after the rust audit findings we agreed to remove the current implementation for the caches and move to a completly lazy cache without on-chain event streaming:
> 
> [0763079](https://github.com/worldcoin/world-id-protocol/commit/076307958ea6fb3d3a2590fa86dc2f0b9ef3749f)
> 
> so this is not needed anymore - we also removed the maintaince tasks completly removing the necessacity for this PR. Would close this if that is fine with you

oh ok, got it yeah I'm closing this

PR #712, opened 2026-05-01
feat: support session proof generation

Takaros999:
This PR adds runtime support for session proof generation on top of the proof type request contract.

- routes create-session and prove-session flows through explicit `proof_type` handling
- derives or re-derives the session randomness needed for proof generation
- validates session IDs before assembling session responses
- scopes OPRF nonce replay tracking across uniqueness, session seed, and session action purposes
- notes that WIP-101 does not currently support session proofs

Validation:
- `cargo check -p world-id-proof --no-default-features`
- `cargo test -p world-id-oprf-node nonce_history --lib`
- `cargo test -p world-id-oprf-node auth::rp_module::tests::test_session_success --lib`
- `cargo check -p world-id-core --no-default-features`

### DIFF
--- a/crates/authenticator/src/prove.rs
+++ b/crates/authenticator/src/prove.rs
@@ -1,7 +1,7 @@
 use secrecy::ExposeSecret;
 use world_id_primitives::{
-    Credential, FieldElement, ProofRequest, ProofResponse, RequestItem, ResponseItem, SessionId,
-    SessionNullifier, ZeroKnowledgeProof,
+    Credential, FieldElement, ProofRequest, ProofResponse, ProofType, RequestItem, ResponseItem,
+    SessionId, SessionNullifier, ZeroKnowledgeProof,
 };
 use world_id_proof::{
     AuthenticatorProofInput, FullOprfOutput, OprfEntrypoint, ProofCompression,
@@ -134,6 +134,7 @@ impl Authenticator {
         proof_request: &ProofRequest,
         account_inclusion_proof: Option<AccountInclusionProof<TREE_DEPTH>>,
     ) -> Result<FullOprfOutput, AuthenticatorError> {
+        proof_request.validate_proof_type()?;
         let mut rng = rand::rngs::OsRng;
 
         let oprf_entrypoint = self.get_oprf_entrypoint(account_inclusion_proof).await?;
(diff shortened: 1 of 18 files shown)


Takaros999:
@codex review

crates/authenticator/src/prove.rs
```diff
@@ -277,28 +292,36 @@ impl Authenticator {
...
         // 2. Resolve session seed
-        let resolved_session_seed = if proof_request.is_session_proof() {
-            if let Some(seed) = session_id_r_seed {
-                // Validate the cached seed produces the expected session ID
+        let (resolved_session_id, resolved_session_seed) = match proof_request.proof_type {
+            ProofType::Uniqueness => (None, None),
+            ProofType::CreateSession => {
+                let (session_id, seed) = self
+                    .build_session_id(proof_request, None, account_inclusion_proof)
+                    .await?;
```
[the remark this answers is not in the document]
Takaros999: Evaluated: I do not think this should change in this PR. For `CreateSession`, `session_id_r_seed` alone is not enough to reproduce a pre-created session because the corresponding `oprf_seed` is part of the `SessionId`, and `CreateSession` intentionally rejects a request `session_id`. The supported flow here is that create-session mints the session inside `generate_proof`; cached `session_id_r_seed` is meaningful for `ProveSession`, where the request carries the existing `SessionId` and therefore the `oprf_seed`. Reusing only the provided seed in create-session would still mint a different session with a fresh `oprf_seed`, so it would not solve the scenario described.

crates/authenticator/src/prove.rs
```diff
@@ -195,33 +197,44 @@ impl Authenticator {
         session_id_r_seed: Option<FieldElement>,
         account_inclusion_proof: Option<AccountInclusionProof<TREE_DEPTH>>,
     ) -> Result<(SessionId, FieldElement), AuthenticatorError> {
+        proof_request.validate_proof_type()?;
```
paolodamico, line 203: This should be part of the initial validation right? IIRC there are other validations on the proof request

paolodamico (commented):
Aren't we missing generating a session proof when a new SessionId gets created ?

services/oprf-node/src/auth/rp_module.rs
```diff
@@ -402,9 +403,20 @@ impl RpModuleAuth {
         .await?;
 
         tracing::trace!("add nonce to store...");
-        // add nonce to history to check if the nonces where only used once
+        // Add nonce to history to check if the nonce was only used once in this scope.
+        let nonce_scope = match self.kind {
```
0xThemis, line 407: If we scope the nonces it is not necessary to build two `nonce_histories` in the lib.rs. Would be nice to also have this in this PR so that we only create a single instance.
https://github.com/worldcoin/world-id-protocol/blob/cc551012e227740c35c4f0c4452688669a6d49a1/services/oprf-node/src/lib.rs#L169

Takaros999: updated

crates/authenticator/src/prove.rs
Takaros999: can you be more specific? I wasn't able to find a better spot

Takaros999:
@codex review

0xThemis (dismissed):

paolodamico (approved):

0xThemis (approved):

PR #729, opened 2026-05-13
refactor!: model indexer/gateway as ServiceEndpoint enum on Config

kilianglas:
## Summary
General goal: Move the OHTTP config into the shared `Config` struct, to make the overal config structure less confusing

- Replaces the flat `indexer_url` / `gateway_url` `String` fields on `Config` with a `ServiceEndpoint` enum that captures the transport choice explicitly: `Direct { url }` or `Ohttp { url, relay_url, key_config_base64 }`. The URL is required in both variants; OHTTP-specific fields only exist in the `Ohttp` variant.
- Removes `AuthenticatorConfig` and its lossy `From<Config>` conversion. `Authenticator::init` / `register` / `init_or_register` now take `Config` directly. This eliminates the seam that silently dropped OHTTP relay configuration on conversion (see walletkit `init_with_defaults` regression).
- `OhttpClientConfig` is retained as the runtime input to `OhttpClient::new`; `ServiceClient::new` now takes `&ServiceEndpoint` and builds the OHTTP client only for the `Ohttp` variant.
- JSON shape on `Config::from_json` becomes internally-tagged: `{ "indexer": { "type": "direct", "url": "..." }, "gateway": { "type": "ohttp", "url": "...", "relay_url": "...", "key_config_base64": "..." } }`. Walletkit's defaults JSON will need to be updated alongside the version bump (tracked separately).

## Breaking changes

- `Config::new` signature: `indexer_url: String, gateway_url: String` -> `indexer: ServiceEndpoint, gateway: ServiceEndpoint`.
- `AuthenticatorConfig` removed. Callers pass `Config` directly to `Authenticator::*`.
- `Config::indexer_url()` / `gateway_url()` now return `&str` (convenience accessors over the variant URL). The full endpoint is available via `Config::indexer()` / `gateway()`.
- `Config` JSON shape changed for the indexer/gateway fields (see Summary).

### DIFF
--- a/crates/authenticator/src/authenticator.rs
+++ b/crates/authenticator/src/authenticator.rs
@@ -16,7 +16,6 @@ use crate::{
     },
     service_client::{ServiceClient, ServiceKind},
 };
-use serde::{Deserialize, Serialize};
 use world_id_primitives::{Credential, FieldElement, ProofResponse, Signer};
 
 pub use crate::ohttp::OhttpClientConfig;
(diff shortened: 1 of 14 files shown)


0xThemis (approved):

alessandromazza98 (approved):
LGTM

PR #740, opened 2026-05-15
fix: num leaves input validation

0xForerunner:


services/indexer/src/tree/state.rs
```diff
@@ -64,6 +64,11 @@ impl TreeState {
         1usize << self.inner.tree_depth
     }
 
+    /// Returns the tree capacity (2^depth).
+    pub async fn num_leaves(&self) -> usize {
+        1usize << self.inner.tree.read().await.num_leaves()
```
0xOsiris: This is not correct

0xForerunner: yeah sorry already pushed a fix. Coppied it over to fast

0xOsiris (approved):

0xOsiris (approved):

0xForerunner:
Will let CI finish then merge

PR #751, opened 2026-05-23
feat: poll blockchain for events instead of websocket streaming

kilianglas:
## Summary

Switches the indexer's blockchain provider from WebSocket subscriptions to periodic HTTP polling.

Previously the indexer backfilled over HTTP and then transitioned to live WebSocket streaming, which required a separate `WS_URL` endpoint and the `provider-ws` alloy feature, and reconnected on every dropped socket. It now syncs entirely over HTTP RPC, polling for new events on a fixed interval.

## Changes

- `blockchain/mod.rs`: replace `backfill_and_stream_events` (websocket) with `pull_events` (periodic HTTP polling)
- Remove `WS_URL` / `ws_rpc_url` config, the `MissingWsUrl` error, and the `provider-ws` alloy feature
- Add `BLOCKCHAIN_POLL_INTERVAL_SECS` (default `1`) to control the poll interval
- Update `.env.example`, README, and integration tests accordingly

---

> [!NOTE]
> **Medium Risk**
> Changes how the indexer ingests chain data (latency, RPC load, and gap handling differ from WS); core commit/reorg logic is unchanged but operational behavior at the sync boundary shifts.
> 
> **Overview**
> Replaces the indexer’s **HTTP backfill + WebSocket live stream** pipeline with **continuous HTTP RPC polling**, so chain sync no longer needs a separate WebSocket endpoint or reconnect logic.
> 
> **`pull_events`** polls the chain head, fetches registry logs in batches up to the latest block, then sleeps for **`BLOCKCHAIN_POLL_INTERVAL_SECS`** (default 1s) before the next round. **`WS_URL`**, **`ws_rpc_url`**, **`MissingWsUrl`**, and the **`provider-ws`** alloy feature are removed; **`Blockchain::new`** is synchronous (HTTP provider only).
> 
> **`process_registry_events`** drives the new stream and still restarts the loop on stream errors, with an explicit **`BLOCKCHAIN_RETRY_DELAY`** sleep. Docs, **`.env.example`**, config/tests, and blockchain integration tests are updated (WS/gap-fill scenarios dropped in favor of poll + RPC failure tests). **`deny.toml`** adds an ignored advisory for **`proc-macro-error2`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba4c9cf22cfbcaad91549260de5f1a6a1bfcfe58. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/indexer/src/blockchain/mod.rs
0xOsiris, line 121: I don't think a stream makes sense here. It'd be more performant to use a `JoinSet` or `FuturesUnordered` and concurrently fetch all logs in batches then sort by key on the block number

services/indexer/src/blockchain/mod.rs
```diff
@@ -107,219 +68,51 @@ impl Blockchain {
...
-            Ok::<_, BlockchainError>(
-                missed_logs
-                    .chain(stream::iter(std::iter::once(first_decoded)))
-                    .chain(ws_stream.map(|log| {
-                        if let Some(block_number) = log.block_number {
-                            crate::metrics::set_chain_head_block(block_number);
+        stream::unfold(from_block, move |current_from| async move {
+            // Poll the chain head until there is a non-empty range to fetch.
+            let latest_block = loop {
+                match self.get_block_number().await {
```
0xOsiris, line 85: This should be transiently handled, as this breaks the whole consumer loop and forces a restart AFAICT

kilianglas: Transient errors are handled by provider layers defined in `common`

services/indexer/src/blockchain/mod.rs
kilianglas: This would certainly be more performant, but this function mainly exists to prevent running into rate limits, and concurrent requests likely don't help here. I'd keep it as it is in this PR, an optimize this later. Note, that this exact function existed before.

0xOsiris (dismissed):

services/indexer/src/blockchain/mod.rs
0xOsiris: Sure

0xOsiris (approved):

services/indexer/src/blockchain/mod.rs
0xOsiris: Again this should be handled at the provider layers, and we should have our own nodes as fallbacks with effectively no concurrency limit

PR #753, opened 2026-05-25
feat: credential fees

0xForerunner:
This WIP adds per-proof fees for World ID credentials. Fees are configured per `issuerSchemaId` and charged only when a proof actually uses a credential for that schema.


kilianglas:
@0xForerunner Why was this closed?

PR #769, opened 2026-06-03
refactor(oprf): info for invalid merkle errors

0xThemis:
- **feat(oprf): print some information on invalid merkle root**
- **refactor(node): cleanup error variants**

### DIFF
--- a/services/oprf-node/src/auth/merkle_watcher.rs
+++ b/services/oprf-node/src/auth/merkle_watcher.rs
@@ -19,7 +19,7 @@ use eyre::Context;
 use moka::future::Cache;
 use taceo_nodes_common::web3;
 use tracing::instrument;
-use world_id_primitives::FieldElement;
+use world_id_primitives::{FieldElement, oprf::WorldIdRequestAuthError};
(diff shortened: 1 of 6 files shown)


services/oprf-node/src/auth/merkle_watcher.rs
```diff
@@ -11,30 +11,46 @@
...
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum MerkleWatcherError {
-    #[error("invalid Merkle root")]
-    InvalidMerkleRoot,
-    #[error("merkle root unknown")]
-    UnknownMerkleRoot,
+    #[error(
+        "invalid Merkle root at block #{block}. Timestamp on block {timestamp_block} - root time stamp from contract: {root_time_stamp}"
+    )]
+    InvalidMerkleRoot {
```
dkales, line 33: I guess we log it somewhere else, but should we not include the invalid Merkle root as well?

dkales (dismissed):

services/oprf-node/src/auth/merkle_watcher.rs
0xThemis: good point - is in span event for ensure valid but will add

dkales (approved):

PR #771, opened 2026-06-04
fix: cleanup indexer log decoding & RPC polling

0xOsiris:


kilianglas:
The improvement this tries to achieve should be mostly covered by PR #748 and PR #751

services/indexer/src/blockchain/events.rs
```diff
@@ -15,73 +28,22 @@ pub struct BlockchainEvent<T: Clone> {
...
-    AuthenticatorInserted(AuthenticatorInsertedEvent),
-    AuthenticatorRemoved(AuthenticatorRemovedEvent),
-    AccountRecovered(AccountRecoveredEvent),
-    RootRecorded(RootRecordedEvent),
-}
-
-impl RegistryEvent {
-    pub fn signatures() -> Vec<FixedBytes<32>> {
+impl RegistryEventExt for RegistryEvent {
+    fn signatures() -> Vec<FixedBytes<32>> {
```
Dzejkop, line 46: Signatures of all events are actually exported as `WorldIdRegistryEvents::SIGNATURES` - it's `&'static [&'static str]` unfortunately but we could easily map that into a `Vec<FixedBytes<32>>`

PR #777, opened 2026-06-09
refactor: change indexer blockchain poll interval unit to milliseconds

kilianglas:
## What

Renames the indexer's `BLOCKCHAIN_POLL_INTERVAL_SECS` env var to `BLOCKCHAIN_POLL_INTERVAL_MS`, switching the head-poll interval from second to ms granularity, to enable smaller poll intervals to reduce latency. Default changes `1` → `1000`, so behaviour is unchanged out of the box.


services/indexer/src/config.rs
```diff
@@ -520,14 +520,14 @@ mod tests {
         set_env("START_BLOCK", "12345");
         set_env("BATCH_SIZE", "128");
         set_env("TREE_MAX_BLOCK_AGE", "500");
-        set_env("BLOCKCHAIN_POLL_INTERVAL_SECS", "5");
+        set_env("BLOCKCHAIN_POLL_INTERVAL_MS", "500");
```
alessandromazza98, line 523: this should be 5000 to be equal

services/indexer/src/config.rs
```diff
@@ -520,14 +520,14 @@ mod tests {
...
 
         let config = IndexerConfig::from_env();
 
         assert_eq!(config.start_block, 12345);
         assert_eq!(config.batch_size, 128);
         assert_eq!(config.tree_max_block_age, 500);
-        assert_eq!(config.blockchain_poll_interval_secs, 5);
+        assert_eq!(config.blockchain_poll_interval_ms, 500);
```
alessandromazza98, line 530: same here

alessandromazza98 (approved):
lgtm, but take care of my comments

Dzejkop (approved):

PR #784, opened 2026-06-11
feat: add per-endpoint rpc transport metrics

kilianglas:
## Summary
Adds a `EndpointMetricsLayer` to the shared RPC provider. 

The layer sits *below* the retry and fallback layer, and records which endpoint was used by a requests (interesting if there mulitple RPC endpoints are configured) via a counter, as well as request latency in a histogram.


services/common/src/provider.rs
```diff
@@ -114,6 +116,18 @@ fn pod_ordinal() -> Option<usize> {
     parse_ordinal(&std::env::var("HOSTNAME").ok()?)
 }
 
+/// Low-cardinality, secret-free metric label for an RPC endpoint:
+/// the host, plus `:port` when the port is non-default for the scheme.
+/// Strip full path to avoid leaking API keys.
```
Dzejkop: I don't know why, but reading this doc comment feel like I'm having a stroke.

```suggestion
/// Strips the url to leave only the host & port
```

services/common/src/provider.rs
```diff
@@ -337,9 +356,11 @@ impl ProviderArgs {
             });
         let retry_layer = RetryLayer::new(retry_policy, &retry_cfg);
 
-        // Flow is: RetryLayer calls ThrottleLayer calls FallbackLayer calls transports
-        // I.e. if throttling is enabled retries count into the request budget
-        // NOTE: Retries can be disabled by setting max_retries to 0 in the retry config. Layer could be made optional as well.
+        // Flow is: RetryLayer calls ThrottleLayer calls FallbackLayer calls
+        // EndpointMetricsLayer calls transports.
+        // I.e. if throttling is enabled retries count into the request budget.
```
Dzejkop:
```suggestion
```

seems unnecessary?

services/common/src/provider.rs
```diff
@@ -337,9 +356,11 @@ impl ProviderArgs {
...
+        // NOTE: Retries can be disabled by setting max_retries to 0 in the retry config.
+        // Layer could be made optional as well in the future.
```
Dzejkop: I know this is copied over, but what does this even mean? Maybe we should remove it?


```suggestion
```

services/common/src/provider.rs
```diff
@@ -315,9 +329,14 @@ impl ProviderArgs {
 
         let num_urls = http.len();
 
-        let transports = http
+        let labeled_transports = http
             .into_iter()
-            .map(|url| Http::with_client(http_client.clone(), url))
+            .map(|url| {
+                // Leaks a static str to avoid per-request metric-label allocs.
```
Dzejkop, line 333: could also be `Arc`? Then we'd avoid cloning?

services/common/src/provider_layers.rs
```diff
@@ -290,3 +290,121 @@ where
...
+        + Clone
+        + Send
+        + Sync
+        + 'static,
+    S::Future: Send,
+{
+    type Response = ResponsePacket;
+    type Error = TransportError;
+    // Must be exactly `TransportFut<'static>`: the `FallbackService` this
+    // feeds into requires `S: Service<RequestPacket, Future = TransportFut<'static>, ...>`.
```
Dzejkop:
```suggestion
```

this would show up as a compiler error if violated

services/common/src/provider_layers.rs
```diff
@@ -290,3 +290,121 @@ where
         })
     }
 }
+
+/// Counter: completed RPC transport calls, labels `endpoint` and `status`
```
Dzejkop, line 294: worth considering moving this stuff into it's own module - and splitting this file into multiple modules one for each layer

kilianglas: Was thinking about this. I think it's still small enough to be a single file, but could also split it if you want. Don't have a super strong opinion

services/common/src/provider.rs
kilianglas: The reqwest client uses an Arc internally, so cloning is not an issue

services/common/src/provider_layers.rs
Dzejkop: yeah, we can keep it like this for now

services/common/src/provider.rs
Dzejkop: I mean putting the endpoint into an `Arc<String>` instead of `String` and leaking into `&'static str`

Dzejkop (approved):

PR #785, opened 2026-06-14
feat: clearer OHTTP client error messages

kilianglas:
## Summary

Improves the accuracy of error messages surfaced on the OHTTP request path in the authenticator. Today, non-crypto failures are misattributed to OHTTP crypto — e.g. a target service being unreachable surfaces as a cryptic "decapsulation" error, and a captive portal returning a 200 HTML page also fails as an opaque decapsulation error.

All changes are server-agnostic and do **not** depend on matching the server's error body (so they remain portable across any RFC 9458 gateway). No server change is required.

## Changes

- **Split decapsulation from encapsulation errors** — `OhttpEncapsulationError` previously printed "encapsulation error" for *both* encrypt and decrypt failures. Added a dedicated `OhttpDecapsulationError` so a failed decrypt is labelled accurately.
- **Content-type guard before decapsulation** — a 2xx relay response that is not `message/ohttp-res` (loose, case-insensitive match) now returns `OhttpRelayInvalidResponse` instead of a cryptic decap failure. The gateway only sets that content-type on success, so this is provably a relay-leg condition — most commonly a captive portal / transparent proxy intercepting the client network.
- **Relay errors carry scope + relay URL** — `OhttpRelayError` now includes the `service` scope and `relay_url`, and its message reflects that a non-2xx outer response means the request was rejected *before* decapsulation (relay leg, or the gateway's RFC 9458 §5.2 pre-decap plaintext rejections such as a 401 key/config mismatch).
- **Inner 502/503/504 → service unavailable** — inner statuses that signal an unreachable/overloaded upstream now map to a clear `ServiceUnavailable` error rather than a raw 5xx body. RFC 9458 §5.2 uses 504 for a gateway that cannot reach its target; this installs the correct client behaviour now (no effect on the current gateway, which returns 500, until a server follow-up lands).

## Out of scope

The headline "target-unreachable shows as decapsulation" symptom is **not** client-fixable: RFC 9458 §5.2 mandates that post-decapsulation and target errors be encapsulated (inner response, outer 200), so there is no outer-layer signal. The portable fix is server-side — emit a distinct inner status (504, not 500) for target-unreachable — and is recommended as a follow-up on the gateway.

## Tests

- Pure unit test for the loose `message/ohttp-res` matcher.
- mockito-based request-path tests: relay non-2xx → `OhttpRelayError` (scope + URL populated); 2xx non-OHTTP content-type → `OhttpRelayInvalidResponse`; correct content-type with an undecryptable body → `OhttpDecapsulationError`.

### DIFF
--- a/crates/authenticator/src/error.rs
+++ b/crates/authenticator/src/error.rs
@@ -98,21 +98,53 @@ pub enum AuthenticatorError {
         max_supported_slot: usize,
     },
 
-    /// OHTTP encapsulation or decapsulation error.
+    /// OHTTP encapsulation error (encrypting the request, or parsing the
+    /// relay's key config list). This is a client-side crypto/encoding failure.
     #[error("OHTTP encapsulation error: {0}")]
     OhttpEncapsulationError(#[from] ohttp::Error),
+
+    /// OHTTP decapsulation error: the relay returned a `message/ohttp-res`
+    /// response that could not be decrypted. Indicates a genuine OHTTP crypto
+    /// failure (e.g. key mismatch), distinct from a non-OHTTP response body.
+    #[error("OHTTP decapsulation error: {0}")]
+    OhttpDecapsulationError(ohttp::Error),
 
     /// Binary HTTP framing error.
     #[error("Binary HTTP error: {0}")]
     BhttpError(#[from] bhttp::Error),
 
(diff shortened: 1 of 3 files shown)


paolodamico:
@codex review

PR #828, opened 2026-06-30
refactor(proof): single provenance per ZK artifact type

Dzejkop:
Each ZK artifact type now has exactly one way of being obtained — no more "download or build" ambiguity:

- **Circom artifacts** (trusted-setup outputs, cannot be rebuilt): committed files when building in-repo, otherwise downloaded from the pinned `circuit-artifacts-*` GitHub release. Downloads are now **verified against SHA-256 digests pinned in build.rs**. Overriding the tag via `WORLD_ID_CIRCUIT_ARTIFACT_RELEASE_TAG` skips verification with a warning (dev escape hatch).
- **Noir ownership proof artifacts** (`ownership_proof.pkp`/`.pkv`): always **built ad-hoc from the checked-in circuit source with `nargo`** when `embed-noir-artifacts` is enabled. The noir download path and the `build-noir-artifacts` feature are removed.

The `generate-noir-artifacts` tool is **dissolved entirely**: build.rs runs `nargo` and the provekit R1CS compiler directly (both provekit crates come from crates.io, so `world-id-proof` stays publishable with no path-only dependencies). To keep ad-hoc builds byte-identical across machines, the build fails fast with an actionable error when `nargo` is missing from PATH or is not the pinned version (`1.0.0-beta.11`, same pin as `flake.nix`).

CI/release changes:
- `release-circuit-artifacts.yml` no longer builds/publishes noir artifacts — the release only carries the Circom files.
- `release-crates.yml` drops the noir download smoke test (there is no noir download path anymore).
- `rust-ci.yml` drops the `build-noir-artifacts` feature flags; building is unconditional when `embed-noir-artifacts` is on and nargo is already installed in those jobs.

Verified locally: default-feature build (local circom files), `embed-noir-artifacts` build with flake-pinned nargo, the actionable error without nargo, and the download+checksum path against the live release (with `circom/` hidden).

Part of a stack with #831 (ZkArtifactSource trait) and #833 (granular embed features).

---

> [!NOTE]
> **Medium Risk**
> Touches ZK build and ownership-proof APIs and default feature wiring; incorrect nargo or artifact handling could break proving compatibility across builders, but changes are localized and builds fail fast on toolchain mismatch.
> 
> **Overview**
> **Each ZK artifact type now has one clear source of truth**, replacing mixed download/build paths and separate prove/verify feature flags.
> 
> **Noir ownership proofs:** `zk-ownership-prove` / `zk-ownership-verify` and the `generate-noir-artifacts` crate are removed. With **`embed-noir-artifacts`**, `build.rs` compiles the checked-in circuit via pinned **`nargo`** (`1.0.0-beta.11`) and provekit, failing fast on missing or wrong toolchain. Runtime APIs add explicit PKP/PKV loaders plus **`generate_ownership_proof_with_prover`** / **`verify_ownership_proof_with_verifier`**; embedded helpers stay behind the feature. The module is **native-only** (`wasm32` excluded); provekit prover/verifier are **target-gated** dependencies.
> 
> **Circom:** Release workflow and docs emphasize **Circom-only** GitHub artifacts (Noir PKP/PKV no longer published there). **`authenticator`** stops enabling a dedicated ownership-prove feature on `world-id-proof`.
> 
> **CI / workspace:** `cargo check --workspace --all-features`, circuit release job simplified (no Rust/Noir build step), **`release-crates`** drops **`noirup`**, **`Justfile`** drops **`build-noir-artifacts`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff1704b96a8273a7bd1e5ce038df5eccbb2ff4db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


paolodamico:
great stuff! I do still see some issues here though:
1. if you're working on a new circuit or updating a circuit, the Rust CI will still run against the latest release artifacts. most circuit changes will also require upstream code changes so this would be a problem.
2. after you push a circuit update (and consequently rust code updates), we may forget to push a circuits release, perhaps there should be another check that if you update the artifacts then a release must occur

kilianglas (dismissed):

Dzejkop:
> great stuff! I do still see some issues here though:
> 
> 1. if you're working on a new circuit or updating a circuit, the Rust CI will still run against the latest release artifacts. most circuit changes will also require upstream code changes so this would be a problem.
> 2. after you push a circuit update (and consequently rust code updates), we may forget to push a circuits release, perhaps there should be another check that if you update the artifacts then a release must occur

after some thinking I decided to go with the following approach:
1. This repo CI always builds the noir circuits ad-hoc in build.rs
2. Upon a release (of crates for now, but I'm going to add a release workflow for services as well) the release PR has CI that uses the download path

this means that:
1. we can work locally using the ad-hoc build path
2. merges to main that affect circuits won't break ci because ci will also use the ad-hoc path
3. we ensure that downstream consumers of released versions of our libraries rely on a crate that can work with the download path - enforced by the release PR CI

crates/authenticator/Cargo.toml
```diff
@@ -70,7 +70,7 @@ reqwest = { workspace = true, features = ["rustls-tls"] }
 tokio = { workspace = true, features = ["rt-multi-thread", "net"] }
 rustls = { workspace = true }
 webpki-roots = { workspace = true }
-world-id-proof = { workspace = true, features = ["zk-ownership-prove"] }
+world-id-proof = { workspace = true }
```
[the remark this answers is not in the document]
Dzejkop: this is a valid concern - but given that we don't expose an alternative way for loading proof material - I think it's fine to leave it like this for now - while I work on a better abstraction layer for proof material loading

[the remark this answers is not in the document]
paolodamico: but if we merge this wouldn't ownership proofs be unusable by authenticators?

tools/generate-noir-artifacts/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "generate-noir-artifacts"
-publish = false
+description = "Generates Noir ownership proof artifacts for World ID"
+publish = true
```
paolodamico: why is this crate published? shouldn't it be a utility?

crates/proof/src/lib.rs
```diff
@@ -24,7 +24,8 @@ use provekit_common::{InputMap, InputValue, NoirElement};
 
 use world_id_primitives::FieldElement;
 
-#[cfg(any(feature = "zk-ownership-prove", feature = "zk-ownership-verify"))]
+// TODO: Currently ownership proofs are not supported for WASM targets
```
paolodamico, line 27: **high**. I'm not sure about this, having feature flags for proving and verification is useful for places where final binary size is a concern. for example, an authenticator doesn't need the `.pkv` artifact because it doesn't need to verify proofs.

.github/workflows/release-crates.yml
```diff
@@ -27,10 +27,8 @@ jobs:
         with:
           toolchain: stable
 
-      - uses: noir-lang/noirup@7dbe69ccc78877f0200ffa5a40836c953d2cfd8f # v0.1.4
-        with:
-          # See https://github.com/worldfnd/provekit, a specific nargo toolchain version is required
-          toolchain: v1.0.0-beta.11
+      - name: Validate circuit artifact download path
```
paolodamico: what is the goal here? we want to make sure that artifacts can be downloaded?

Dzejkop: yes and that the correct circuit artifacts have been released and referenced in code

crates/proof/src/lib.rs
Dzejkop: we can have a more granular setup with a feature flag per embedded artifact

tools/generate-noir-artifacts/Cargo.toml
Dzejkop: code is reused between world-id-proof's build.rs and CI AFAIK in order to use it as a build dep of world-id-proof we'll need to publish it

crates/authenticator/Cargo.toml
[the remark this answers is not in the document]
Dzejkop: not quite, the downstream authenticator crate can specify the feature flags for world-id-proof enabling artifact embedding

crates/proof/src/lib.rs
Dzejkop: I'll have that setup in a future PR

Dzejkop: resolved in https://github.com/worldcoin/world-id-protocol/pull/833

crates/proof/src/ownership_proof.rs
Dzejkop, line 1: note - the changes in this file are kinda messy, but will be cleaned up a lot in https://github.com/worldcoin/world-id-protocol/pull/833

kilianglas (approved):

PR #830, opened 2026-06-30
feat: billing contract v1 (WIP-107)

kilianglas:
## Summary

First version of the on-chain **Billing Contract** (WIP-107): OPRF nodes count unique RP requests per ~epoch, submit EIP-712-signed votes, and the contract finalizes the lower-median count (gated on quorum), prices it through a tiered WLD fee schedule, accrues per-RP debt, and surfaces `is_blocked` for RPs past their payment window.

Scope of this PR: **contract + `IBillingContract` interface + Foundry tests only**. No deploy script and no OPRF node-side Rust (separate follow-up entries).

## Deviations from initial spec

* Voting is chunked to allow for larger numbers of RPs
* Finalization of and epoch is an async/ separate operation

### DIFF
--- a/contracts/src/core/BillingContract.sol
+++ b/contracts/src/core/BillingContract.sol
@@ -417,6 +417,20 @@ contract BillingContract is WorldIDBase, IBillingContract {
         return uint64(end);
     }
 
+    /// @notice The timestamp at which epoch `epoch`'s voting window closes.
+    /// @dev Voting is open over `[epochEnd(epoch), votingWindowEnd(epoch))`. The window size is
+    ///      governed by the era of epoch `epoch + 1` (see {_votingWindowEraOf}), so this stays exact
+    ///      across timing changes rather than relying on the current era.
+    function votingWindowEnd(uint32 epoch) public view virtual returns (uint64) {
+        return epochEnd(epoch) + _votingWindowEraOf(epoch).votingWindow;
+    }
+
+    /// @notice The timestamp at which epoch `epoch`'s payment window closes; payment is overdue after it.
+    /// @dev End of the voting window plus the payment window, both governed by the era of epoch `epoch + 1`.
+    function paymentWindowEnd(uint32 epoch) public view virtual returns (uint64) {
+        return _paymentDue(epoch);
+    }
+
     /// @notice The OPRF key registry address the node set is read from.
     function getOprfKeyRegistry() external view virtual onlyProxy onlyInitialized returns (address) {
         return address(_oprfKeyRegistry);
--- a/contracts/src/core/BillingContract.sol
+++ b/contracts/src/core/BillingContract.sol
@@ -27,9 +27,11 @@ interface IOprfNodeSet {
  * @dev OPRF nodes count unique RP requests per epoch and submit EIP-712-signed votes. The contract
  *      finalizes the lower-median count across the node set (gated on quorum), prices it through a
  *      tiered WLD fee schedule, accrues per-RP debt, and blocks RPs past their payment window.
- *      Finalization is permissionless and chunkable ({finalizeEpochs}); {submitBillingVotes} and
- *      {pay} opportunistically flush a bounded chunk. Finalizing prunes raw vote data, so raw state
- *      scales with participants, not time, and no single call can be bricked by an oversized epoch.
+ *      Finalization is permissionless and chunkable ({finalizeEpochs}) — it is the sole driver, so
+ *      a keeper must run it (cadence < paymentWindow keeps {is_blocked} never-late). Neither
+ *      {submitBillingVotes} nor {pay} finalize as a side effect, keeping their gas predictable.
+ *      Finalizing prunes raw vote data, so raw state scales with participants, not time, and no
+ *      single call can be bricked by an oversized epoch.
  * @custom:repo https://github.com/worldcoin/world-id-protocol
  */
 contract BillingContract is WorldIDBase, IBillingContract {
@@ -120,11 +122,6 @@ contract BillingContract is WorldIDBase, IBillingContract {
     /// @dev EIP-712 typehash for a single RpCount struct (member of a BillingVote).
     bytes32 public constant RPCOUNT_TYPEHASH = keccak256("RpCount(uint64 rpId,uint64 count)");
 
-    /// @dev Finalization budget for the opportunistic flush done by `submitBillingVotes`/`pay`.
-    ///      One unit per RP finalized and one per epoch closed; permissionless `finalizeEpochs`
-    ///      takes an explicit budget for bulk catch-up.
-    uint256 internal constant DEFAULT_FINALIZE_STEPS = 256;
-
     ////////////////////////////////////////////////////////////
     //                        Constructor                     //
     ////////////////////////////////////////////////////////////
@@ -198,9 +195,9 @@ contract BillingContract is WorldIDBase, IBillingContract {
         if (block.timestamp < votingStart) revert VotingWindowNotOpen();
         if (block.timestamp >= votingStart + _votingWindow) revert VotingWindowClosed();
 
-        // Opportunistically finalize a bounded chunk of now-closed epochs so the system
-        // self-drives without a dedicated keeper. Bounded so a large epoch can never brick submit.
-        _finalize(type(uint64).max, DEFAULT_FINALIZE_STEPS);
+        // No finalization side effect here: submit only records votes. Finalization is driven
+        // solely by the permissionless {finalizeEpochs} (see IBillingContract), so a node's vote
+        // gas is predictable and never carries another epoch's finalization cost.
 
         // Snapshot the node count and fee-schedule version on the epoch's first accepted vote, so
         // quorum/median and pricing stay stable across the voting window even if the live node set
@@ -222,8 +219,23 @@ contract BillingContract is WorldIDBase, IBillingContract {
 
     /// @inheritdoc IBillingContract
     function pay(RpPayment[] calldata payments) external virtual onlyProxy onlyInitialized {
-        // Implemented in "Payment + blocking" step.
-        revert NotImplemented();
+        // Settles currently-finalized debt only; does not finalize (keeper-driven finalizeEpochs
+        // keeps debt current). RPs with no finalized debt are skipped so a batched, permissionless
+        // call is not griefable by a debt that was concurrently settled or never accrued.
+        uint256 len = payments.length;
+        for (uint256 i = 0; i < len; i++) {
+            uint64 rpId = payments[i].rpId;
+            uint256 debt = _totalOwed[rpId];
+            if (debt == 0) continue;
+            if (debt > payments[i].maxAmount) revert DebtExceedsMax();
+
+            // Effects before interaction: clear debt/clock before pulling tokens.
+            _totalOwed[rpId] = 0;
+            _oldestUnpaidDue[rpId] = 0;
+            _feeToken.safeTransferFrom(msg.sender, _feeRecipient, debt);
+
+            emit DebtPaid(rpId, msg.sender, debt);
+        }
     }
 
     /// @inheritdoc IBillingContract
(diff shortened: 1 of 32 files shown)


contracts/src/core/BillingContract.sol
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+    /// @inheritdoc IBillingContract
+    function DOMAIN_SEPARATOR() external view virtual onlyProxy onlyInitialized returns (bytes32) {
+        return _domainSeparatorV4();
+    }
+
+    /// @notice The timestamp at which epoch `epoch` ends (and its voting window opens).
+    function epochEnd(uint32 epoch) public view virtual returns (uint64) {
+        return _genesis + (uint64(epoch) + 1) * _epochLength;
```
0xThemis: Something that came up during implentation on the node side - what happens if `epochLength` changes? 

All past epochs are affected then and the current epoch might change to an epoch that we already voted for.

contracts/src/core/BillingContract.sol
```diff
@@ -0,0 +1,647 @@
...
+    struct PeriodState {
+        uint32 index;
+        uint224 count;
+    }
+
+    ////////////////////////////////////////////////////////////
+    //                        Members                         //
+    ////////////////////////////////////////////////////////////
+
+    // DO NOT REORDER! To ensure compatibility between upgrades, it is exceedingly important
```
alessandromazza98, line 73: question: should we leave a `gap` variable for future updates of this contract?

contracts/src/core/BillingContract.sol
kilianglas: Good point. I'd say it's likely the easiest to add setting `_genesis` in `setTiming` as well. I also think that we can live with losing one or two billing epochs on updates and not have sophisitacted logic to deal with this.

kilianglas: Tbh, we could also just remove most of the setters and just upgrade the contract.

contracts/src/core/interfaces/IBillingContract.sol
```diff
@@ -0,0 +1,307 @@
...
+
+    /**
+     * @notice Settle finalized debt for one or more RPs through caller-selected epochs.
+     * @dev Permissionless; settles currently-finalized debt and does not finalize as a side effect
+     *      (call {finalizeEpochs} first if closed epochs must be reflected). All-or-nothing per
+     *      instruction: pays the debt through `uptoEpoch` or reverts if it exceeds the RP's
+     *      `maxAmount` guard. Pulls WLD from msg.sender to the fee recipient.
+     * @param payments The per-RP payment instructions.
+     */
+    function pay(RpPayment[] calldata payments) external;
```
Kemperino, line 215: nit. i feel like collectDebt would be better naming considering all the accounting logic happening on this call?

contracts/src/core/BillingContract.sol
Kemperino: doesn't this only make sense if another contracts inherits from BillingContract?

contracts/src/core/interfaces/IBillingContract.sol
```diff
@@ -0,0 +1,304 @@
...
+    /// @notice One chunk of an OPRF node's signed billing vote for a single epoch.
+    /// @dev The signed payload is the EIP-712
+    ///      `BillingVoteChunk(uint32 epoch,uint32 chunkIndex,bool isFinal,RpCount[] counts)` struct,
+    ///      where `epoch` is supplied as the call argument shared by every chunk in the batch.
+    struct SignedVoteChunk {
+        // the zero-based chunk index for this node and epoch. Chunks must be submitted in order.
+        uint32 chunkIndex;
+        // true on the final chunk; the node counts as a voter only after this chunk is accepted.
+        bool isFinal;
+        // the per-RP counts reported by the node, strictly ascending globally across chunks,
```
Kemperino, line 37:
```suggestion
        // the per-RP counts reported by the node, strictly ascending by rpId globally across chunks,
```

specifies based on rpId.

contracts/src/core/BillingContract.sol
```diff
@@ -0,0 +1,735 @@
...
+    // clash will occur (effectively a memory safety error).
+
+    /// @dev OPRF key registry the live node set (n, peer addresses) is read from.
+    IOprfNodeSet internal _oprfKeyRegistry;
+
+    /// @dev Append-only history of timing eras; the last entry is the current era. Every epoch
+    ///      that ever existed keeps its true boundaries and windows: epoch `e` is governed by the
+    ///      newest era with `startEpoch <= e`, so timing changes never affect existing epochs.
+    ///      Grows by one entry per {setTiming} call (rare, owner-only), never per epoch.
+    Era[] internal _eras;
```
Kemperino: worth discussing to change to something like the following, then we have future compatibility to change the Era struct.

mapping(uint32 => Era) internal _eras;
uint32 internal _eraCount;

contracts/src/core/BillingContract.sol
```diff
@@ -0,0 +1,735 @@
...
+        uint32 rebatePeriodEpochs
+    ) public virtual initializer {
+        if (oprfKeyRegistry == address(0) || feeToken == address(0) || feeRecipient == address(0)) {
+            revert ZeroAddress();
+        }
+
+        // registrationFee is unused (0); _feeToken (WLD) and _feeRecipient are reused for payments.
+        __BaseUpgradeable_init(EIP712_NAME, EIP712_VERSION, feeRecipient, feeToken, 0);
+
+        _oprfKeyRegistry = IOprfNodeSet(oprfKeyRegistry);
```
Kemperino, line 177: lets fulfill `ERC-165 (supportsInterface) ` if possible on the target oprf registry.

contracts/src/core/interfaces/IBillingContract.sol
```diff
@@ -0,0 +1,324 @@
...
+    error PeriodStateOverflow();
+
+    ////////////////////////////////////////////////////////////
+    //                        EVENTS                          //
+    ////////////////////////////////////////////////////////////
+
+    /// @notice Emitted once per `submitBillingVotes` call.
+    /// @param epoch The epoch the vote chunks were cast for.
+    /// @param count The number of chunks accepted in the call.
+    event VoteChunksSubmitted(uint32 indexed epoch, uint256 count);
```
Kemperino: is this count really useful in any way? i'd argue if we emit somethign the voter itself is more useful than the pure count?

contracts/src/core/BillingContract.sol
```diff
@@ -0,0 +1,735 @@
...
+                epochLength: epochLength,
+                votingWindow: votingWindow,
+                paymentWindow: paymentWindow
+            })
+        );
+        emit TimingUpdated(epochLength, votingWindow, paymentWindow, uint32(eCur) + 1, startTime);
+    }
+
+    /// @inheritdoc IBillingContract
+    function updateOprfKeyRegistry(address newOprfKeyRegistry) external virtual onlyOwner onlyProxy onlyInitialized {
```
Kemperino, line 334: once it has ERC165 support, lets add this here as well

contracts/src/core/BillingContract.sol
```diff
@@ -0,0 +1,735 @@
...
+        _rebatePeriodEpochs = rebatePeriodEpochs;
+        emit RebatePeriodUpdated(rebatePeriodEpochs);
+    }
+
+    ////////////////////////////////////////////////////////////
+    //                    VIEW FUNCTIONS                      //
+    ////////////////////////////////////////////////////////////
+
+    /// @inheritdoc IBillingContract
+    function is_blocked(uint64 rpId) external view virtual onlyProxy onlyInitialized returns (bool) {
```
Kemperino:
```suggestion
    function isBlocked(uint64 rpId) external view virtual onlyProxy onlyInitialized returns (bool) {
```

Kemperino (dismissed):

contracts/src/core/interfaces/IBillingContract.sol
kilianglas: Removed. Event was useless.

contracts/src/core/BillingContract.sol
kilianglas: Fixed.

contracts/src/core/interfaces/IBillingContract.sol
kilianglas: Why? `pay` seems to be the correct name here.

contracts/src/core/BillingContract.sol
kilianglas: I feel like if we need to update this, we can just deploy a new contract version. Not sure if taking on extra complexity here is worth it.

contracts/src/core/interfaces/IBillingContract.sol
```diff
@@ -0,0 +1,360 @@
...
+     * @dev Flushes the tail when node votes have stopped, so debt and `isBlocked` stay current.
+     *      Chunkable: advances the global cursor by at most `maxSteps` units (one per RP finalized,
+     *      one per epoch closed), resuming mid-epoch across calls so a large epoch never bricks.
+     * @param uptoEpoch The highest epoch to finalize up to; capped at the latest closed epoch.
+     * @param maxSteps The maximum units of finalization work to perform in this call.
+     */
+    function finalizeEpochs(uint32 uptoEpoch, uint256 maxSteps) external;
+
+    /**
+     * @notice Replace the tier schedule. Applies to every not-yet-finalized epoch (no versioning).
```
Dzejkop, line 228: should tiering also be era controlled?

contracts/src/core/BillingContract.sol
```diff
@@ -0,0 +1,764 @@
...
+    TimingEra[] internal _timingEras;
+
+    /// @dev Lowest epoch not yet finalized (global finalization cursor).
+    uint32 internal _nextEpochToFinalize;
+
+    /// @dev Number of epochs in a rebate (volume-discount) period.
+    uint32 internal _rebatePeriodEpochs;
+
+    /// @dev The single, current ordered tier schedule (no versioning).
+    Tier[] internal _tierSchedule;
```
Dzejkop, line 90: Same concern as https://github.com/worldcoin/world-id-protocol/pull/830/changes#r3551022582

feels like tiers should also belong to an era

Dzejkop: nevermind, I think it's fine

contracts/src/core/interfaces/IBillingContract.sol
```diff
@@ -0,0 +1,360 @@
...
+     * @dev Owner only. Rates must be strictly decreasing, `upTo` strictly ascending, last `upTo`
+     *      equal to type(uint256).max.
+     * @param tiers The new tier schedule.
+     */
+    function setTierSchedule(Tier[] calldata tiers) external;
```
Dzejkop, line 233: I'm wondering if this shouldn't be in something like `ITieredRebaedBillingContract.sol`

Dzejkop: same for `setRebatePeriodEpochs`

Dzejkop: my point here is that we should leave in `IBillingContract` only the parts relevant to the core billing protocol - i.e. request counting, voting, etc.

how exactly fees are calculated is an "implementation detail" that's not relevant to e.g. oprf nodes

contracts/src/core/interfaces/IBillingContract.sol
```diff
@@ -0,0 +1,360 @@
...
+     *      the end of the in-flight epoch: the new parameters govern later epochs only, and no
+     *      existing epoch is affected in any way — boundaries, open voting windows, and payment
+     *      deadlines all keep their original, era-true values (the full era history is retained).
+     *      If the previous change's era has not started yet, its parameters are updated in place
+     *      instead. Closed-but-unfinalized epochs finalize normally; a change never delays them.
+     * @param epochLength The epoch length in seconds.
+     * @param votingWindow The voting window in seconds.
+     * @param paymentWindow The payment window in seconds.
+     */
+    function setTiming(uint64 epochLength, uint64 votingWindow, uint64 paymentWindow) external;
```
Dzejkop, line 247: imo setter methods in general don't belong in this interface

PR #837, opened 2026-07-06
feat(accountant): skeleton for accountant binary

0xThemis:
- **feat(accountant): skeleton for oprf-accountant binary**
- **chore: add github code owners for accountant**

---

> [!NOTE]
> **Medium Risk**
> New service will store RP signature metadata used for later verification; the ingest path does not yet validate signatures or surface DB errors to clients, so bad or dropped data could go unnoticed until follow-up work.
> 
> **Overview**
> Introduces a new **`services/oprf-accountant`** workspace crate and binary, wired into the root `Cargo.toml` / `Cargo.lock`, with **CODEOWNERS** for the new path.
> 
> The service boots like other OPRF node binaries: **`TACEO_OPRF_ACCOUNTANT__*`** env config (Postgres + bind address default **`0.0.0.0:4322`**), telemetry/metrics hooks, graceful shutdown, and shared **`taceo-nodes-common`** health/version routes merged with an Axum router.
> 
> **Persistence:** SQLx migrations create **`rp_signatures`** (RP id, epoch, nonce, signed validity window, secp256k1 signature bytes) with a unique constraint on **`(rp_id, nonce, epoch)`** for replay protection.
> 
> **API:** **`POST /req`** accepts a JSON batch of **`RpVote`** records and bulk-inserts into Postgres with retries; conflicts are ignored via **`ON CONFLICT DO NOTHING`**. This is explicitly scaffold-level: **`action`** is in the request type but not stored, **`describe_metrics`** is empty, and the handler currently returns **200 OK** even when persistence fails.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1b9193e953d45ec802b344c67c0661e354e137f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/services/oprf-accountant/migrations/20260703123015_init_db.up.sql
+++ b/services/oprf-accountant/migrations/20260703123015_init_db.up.sql
@@ -1,0 +1,19 @@
+-- Stores RP (relying party) signatures over proof requests so they can be verified later.
+create table if not exists rp_signatures (
+    id BIGINT GENERATED ALWAYS AS identity PRIMARY KEY,
+    rp_id BIGINT NOT NULL,
+    epoch BIGINT NOT NULL,
+    -- version is currently always 1
+    version SMALLINT NOT NULL DEFAULT 1,
+    nonce BYTEA NOT NULL,
+    -- Signed validity window (u64 unix seconds stored as bigint)
+    signed_created_at BIGINT NOT NULL,
+    signed_expires_at BIGINT NOT NULL,
+    -- RP ECDSA (secp256k1) signature over the message (alloy Signature, 65 bytes)
+    signature BYTEA NOT NULL,
+    -- Record insertion timestamp (bookkeeping)
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+
+    -- A given nonce should be signed at most once per RP (replay protection)
+    constraint uq_rp_signatures_rp_id_nonce unique (rp_id, nonce, epoch)
+);
--- a/services/oprf-accountant/src/lib.rs
+++ b/services/oprf-accountant/src/lib.rs
@@ -1,0 +1,74 @@
+// #![deny(clippy::all, clippy::pedantic)]
+// #![deny(
+//     clippy::allow_attributes_without_reason,
+//     clippy::assertions_on_result_states,
+//     clippy::dbg_macro,
+//     clippy::decimal_literal_representation,
+//     clippy::exhaustive_enums,
+//     clippy::exhaustive_structs,
+//     clippy::iter_over_hash_type,
+//     clippy::let_underscore_must_use,
+//     clippy::missing_assert_message,
+//     clippy::print_stderr,
+//     clippy::print_stdout,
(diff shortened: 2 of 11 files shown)


Dzejkop (approved):

kilianglas (approved):

fabian1409 (approved):

services/oprf-accountant/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+// #![deny(clippy::all, clippy::pedantic)]
+// #![deny(
```
paolodamico, line 2: nit. what if we start with workspace level lint rules?

services/oprf-accountant/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+//     clippy::missing_assert_message,
+//     clippy::print_stderr,
+//     clippy::print_stdout,
+//     clippy::undocumented_unsafe_blocks,
+//     clippy::unnecessary_safety_comment,
+//     clippy::unwrap_used
+// )]
+// #![allow(missing_docs, reason = "scaffold crate, docs not written yet")]
+
+//! This crate implements the OPRF accountant for World ID.
```
paolodamico, line 20: for later. let's add more details here, and it should definitely reference the spec once available.

services/oprf-accountant/migrations/20260703123015_init_db.up.sql
```diff
@@ -0,0 +1,19 @@
...
+    rp_id BIGINT NOT NULL,
+    epoch BIGINT NOT NULL,
+    -- version is currently always 1
+    version SMALLINT NOT NULL DEFAULT 1,
+    nonce BYTEA NOT NULL,
+    -- Signed validity window (u64 unix seconds stored as bigint)
+    signed_created_at BIGINT NOT NULL,
+    signed_expires_at BIGINT NOT NULL,
+    -- RP ECDSA (secp256k1) signature over the message (alloy Signature, 65 bytes)
+    signature BYTEA NOT NULL,
```
paolodamico, line 13: curious why are we storing the signatures now as the first version relies on trust on this accountant service

paolodamico (approved):

services/oprf-accountant/src/lib.rs
0xThemis: Yeah would be down for that.
We have those lints enabled in our production grade software, but we didn't want to enforce them to the whole repo.

services/oprf-accountant/migrations/20260703123015_init_db.up.sql
0xThemis: In a follow-up we want to have a route so that RPs can fetch their requests from the accountant service so that they can verify which requests were counted. We store the signature now already in preperation for that. Won't be doing anything with that in the first version though.

Everything is subject to change though so if there is a reason to not do it, we can also remove it.

PR #841, opened 2026-07-07
chore: update serial_test to 3.5.0 (RUSTSEC-2026-0205)

kilianglas:
Bumps `serial_test` 3.3.1 -> 3.5.0 in `Cargo.lock`, which drops the transitive `scc 2.4.0` (and `sdd`) dependency flagged today by [RUSTSEC-2026-0205](https://rustsec.org/advisories/RUSTSEC-2026-0205) (unsound `Array::insert`, potential double-free). `scc` only entered the tree via `serial_test`, a dev-dependency of `world-id-indexer`.

Unblocks the `Cargo deny (advisories)` CI check, which currently fails on every branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8927,15 +8927,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "scc"
-version = "2.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


Dzejkop:
fixed in https://github.com/worldcoin/world-id-protocol/pull/840

PR #843, opened 2026-07-07
feat(node): add worker that sends requests to accountant

0xThemis:
- **feat(node): add worker that sends requests to accountant**
- **test(oprf-node): cleanup tests**

---

> [!NOTE]
> **Medium Risk**
> Touches billing/reporting for uniqueness OPRF traffic; failed accountant flushes or full channels can lose billable events, and epoch-less DB writes are incomplete until the TODO is resolved.
> 
> **Overview**
> Adds an **accountant batcher** on the OPRF node that collects billable uniqueness (nullifier) requests after successful RP-module auth and **POSTs batched JSON** to the OPRF-accountant. Flushes run on buffer size, a periodic interval, and graceful shutdown, with **exponential retry**; batches are **dropped** if the accountant stays unreachable after retries, and individual events can be dropped when the worker channel is full (metric added).
> 
> Wires the batcher through **`world-id-oprf-node` startup** (`TACEO_OPRF_NODE__BATCHER__ENDPOINT`, HTTP client timeout) and **`start(..., accountant_batcher)`**; local setup and test stubs use a **`dev_null`** handle where no accountant is needed.
> 
> Refactors **oprf-accountant** HTTP types into `api::BillableRpRequest` (mapped from `NullifierOprfRequestAuthV1`, optional `alloy` signature). Postgres batch insert still expects **epochs**, but the handler **stores empty epoch vectors** until contract lookup is implemented (TODO). RP module tests are consolidated with a shared macro and `RpModuleAuthArgs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9bc541a497cc504bb47f504f3d18c8bbb6969d18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/oprf-node/src/accountant_batcher.rs
```diff
@@ -0,0 +1,350 @@
+//! This file creates an actor that batches [`BillableRpRequest`]s according to WIP-107 and sends them to the `OPRF-accountant` via the configured endpoint.
+//!
+//! All uniqueness proof requests are billable requests in the protocol. Therefore the `OPRF-node` collects all those requests and blindly `POSTs` them to the `OPRF-accountant`. This service is lightweight by choice. It does not perform any form of deduplication on requests (on replicas across the globe). This service expects the `OPRF-accountant` to provide an idempotent endpoint and even sending the same batch multiple times shall gracefully be handled by the `OPRF-accountant`.
+//!
+//! The worker has an internal buffer. If either the buffer reaches its capacity OR a pre-defined flush interval passes, the worker `POSTs` the currently buffered requests to the `OPRF-accountant`. The worker can be configured with a [`backon::ExponentialBackoff`](https://docs.rs/backon/latest/backon/struct.ExponentialBuilder.html). On error cases (after retry), the worker will drop the current batch. This batch won't be reported to the accountant then.
```
fabian1409, line 5: I would note here and/or on the `flush_interval` field that this needs to be lower than the `voting_window` so that the request get forwarded in time

services/oprf-node/src/accountant_batcher.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// The endpoint of the `OPRF-accountant`.
+    ///
+    /// The worker will send `POST` requests to this `URL`.
+    pub endpoint: reqwest::Url,
+
+    /// The max capacity of the internal buffer.
+    ///
+    /// If the buffer reaches this capacity, the worker will `POST` the buffer to the configured endpoint.
+    #[serde(default = "AccountantBatcherConfig::default_buffer_size")]
+    pub buffer_size: NonZeroUsize,
```
fabian1409: nit: maybe name it `buffer_max_size` or `buffer_capacity`

especially in the `AccrountantBatcher` struct right next to the `buffer` field it looks more like we keep track of the length explicitly instead of the max size

services/oprf-accountant/src/lib.rs
```diff
@@ -46,29 +39,6 @@ impl FromRef<AppState> for PostgresDb {
...
-) -> impl IntoResponse {
-    let _ = db.store_request_batch(rp_requests).await;
-    StatusCode::OK
-}
-
 pub async fn start(_config: &OprfAccountantConfig, db: PostgresDb) -> Router {
-    Router::new()
-        .route("/req", post(post_request))
-        .with_state(AppState(db))
+    Router::new().merge(api::routes()).with_state(AppState(db))
```
dkales, line 43: we probably want some kind of body limit

services/oprf-node/src/bin/world-id-oprf-node.rs
```diff
@@ -31,6 +34,12 @@ struct FullWorldOprfNodeConfig {
     /// The postgres config for the secret-manager
     #[serde(rename = "postgres")]
     pub postgres_config: PostgresConfig,
+    /// The config for the accountant batcher worker
+    #[serde(rename = "batcher")]
+    pub accountant_batcher_config: AccountantBatcherConfig,
```
dkales, line 39: Do we want to make this optional? Might help in some local test setups?

services/oprf-node/src/accountant_batcher.rs
```diff
@@ -0,0 +1,350 @@
...
+    buffer_size: usize,
+    rx: mpsc::UnboundedReceiver<AccountantBatcherJob>,
+    flush_task: tokio::task::JoinHandle<()>,
+}
+
+/// Handle to worker.
+///
+/// To initially create a worker with an associated handle, see [`init`]. The handle can be cheaply cloned to get access to the underlying worker. Usually you only want to call `init` once per process.
+#[derive(Debug, Clone)]
+pub struct AccountantBatcherHandle(mpsc::UnboundedSender<AccountantBatcherJob>);
```
dkales: Would not have this unbounded. Large queue size is fine, but not unbounded IMO.
If the submission cannot keep up with incoming queries, it could lead to memory exhaustion and in some way it would get dropped anyways.

dkales (dismissed):

services/oprf-accountant/src/lib.rs
0xThemis: All acountant stuff is still skeleton - but we'll keep that in mind

services/oprf-node/src/bin/world-id-oprf-node.rs
0xThemis: Dont think so for binary tests we want to have the accountant anyways and for integration/unit tests there is the dev_null batcher.

Also then we would need to use `Option` everywhere

services/oprf-node/src/accountant_batcher.rs
0xThemis: Very against bounded channel but happy to discuss.

We have very large backoff logic and if the accountant is down this would mean the whole request flow would block as soon as we hit the limit.

When ignoring wip101 aux data (which is a problem already not quite sure how we want to handle that), we would need more than 4 million requests until we exhaust memory if 1GB.

alessandromazza98 (dismissed):

alessandromazza98 (approved):

dkales (approved):

PR #849, opened 2026-07-09
feat: continue implementing oprf-accountant

fabian1409:
> [!NOTE]
> **Medium Risk**
> Changes affect billing epoch assignment and vote timing against a live contract and wallet key, but votes are not yet submitted on-chain while the cursor can still advance after the stub vote path.
> 
> **Overview**
> Adds an **`OprfAccountantService`** background worker that buckets incoming billable RP requests by epoch (from `expires_at` and on-chain **timing eras**), persists them in Postgres, and on each tick refreshes eras from the **`BillingContract`**, decides whether the last closed epoch’s voting window is open (including a configurable flush **offset**), aggregates per-RP counts, and advances an **`epoch_cursor`** so restarts do not re-vote the same epoch. On-chain **`submitBillingVotes`** / EIP-712 signing is still a **TODO** in `vote_for_epoch`.
> 
> The HTTP **`POST /req`** path now records batches through the accountant instead of writing Postgres directly. **Config** gains billing contract address, RPC, wallet key, tick interval, and voting-window offset; startup spawns the worker alongside Axum and adds bounded graceful shutdown. **Migrations** add **`epoch_cursor`**, store **epoch** on **`rp_signatures`**, and make **signature** optional. **Tests** cover era/epoch math, vote-window edge cases, and Postgres aggregation (including per-epoch nonce uniqueness).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de5c9fe6972934541ac335b27ca883eceaab697c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,905 @@
...
+        // Burst would also be fine, but it is not needed here, Skip would be bad because it could skip epochs
+        interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+
+        loop {
+            tokio::select! {
+                _ = interval.tick() => {
+                    if let Err(err) = self.tick().await {
+                        tracing::error!(error = ?err, "accountant tick failed; retrying next tick");
+                    }
+                }
```
fabian1409, line 275: we could also die instead of log and continue?

services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,905 @@
...
+    /// voting window).
+    ///
+    /// # Errors
+    /// Returns an error if checking the vote window, submitting the vote, or updating the epoch
+    /// cursor fails.
+    #[instrument(level = "info", skip_all, name = "accountant_service::tick")]
+    pub async fn tick(&self) -> eyre::Result<()> {
+        let now = now_unix_timestamp();
+        match self.get_votes_for_current_vote_window(now).await? {
+            VoteWindowResult::NotOpen | VoteWindowResult::AlreadyClosed => return Ok(()),
```
fabian1409: could update the cursor for `AlreadyClosed` and but for now the `get_votes_for_current_vote_window` does not need to check the cursor.
updating it in `AlreadyClosed` would only be useful if we also check it and return early

services/oprf-accountant/migrations/20260703123015_init_db.up.sql
```diff
@@ -10,10 +10,21 @@ create table if not exists rp_signatures (
...
     -- A given nonce should be signed at most once per RP (replay protection)
     constraint uq_rp_signatures_rp_id_nonce unique (rp_id, nonce, epoch)
-);
\ No newline at end of file
+);
+
+-- Singleton table holding the last epoch the accountant has fully processed (i.e. voted
+-- on). `id` is pinned to `true` so there can only ever be one row. Seeded at -1 (no epoch
+-- processed yet) so the first tick starts from epoch 0.
+create table if not exists epoch_cursor (
```
0xThemis: Capitialize to be consistent

services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,905 @@
...
+    timing_eras: Arc<Mutex<Vec<TimingEra>>>,
+    submit_interval: Duration,
+    voting_window_offset: Duration,
+    cancellation_token: CancellationToken,
+}
+
+/// Construction parameters for [`OprfAccountantService::new`].
+pub struct OprfAccountantServiceArgs {
+    /// Provider used to call and sign transactions against the `BillingContract`.
+    pub provider: DynProvider,
```
0xThemis: should we use the provider from web3 nodes_common?

services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,905 @@
...
+    pub cancellation_token: CancellationToken,
+}
+
+impl OprfAccountantService {
+    /// Constructs the service: builds the `BillingContract` binding and validates
+    /// `submit_interval`/`voting_window_offset` against the current timing era.
+    ///
+    /// # Errors
+    /// Returns an error if the current era's timing parameters fail [`validate_timing`].
+    pub async fn new(
```
0xThemis: add panic section that this will panic if the timing eras Vec is empty

services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,905 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+/// Validates `submit_interval` and `voting_window_offset` against `era`'s timing parameters.
+///
+/// # Errors
+/// Returns an error if `era.votingWindow` is not at most `era.epochLength` (consecutive epochs'
```
0xThemis, line 431: Is this a problem if the votingWindows overlap? Cant this just happen or where does this invariant come from?

services/oprf-accountant/src/accountant_service.rs
0xThemis: nah think this is fine so

services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,905 @@
...
+    pub async fn record_rp_request_batch(
+        &self,
+        rp_requests: Vec<BillableRpRequest>,
+    ) -> eyre::Result<()> {
+        tracing::trace!(
+            num_requests = rp_requests.len(),
+            "recording RP request batch"
+        );
+        let timing_eras = self.timing_eras.lock().expect("not poisoned").clone();
+        self.db
```
0xThemis, line 250: nit: I think it is weird that we the accountant calls the db pool and the db pool computes the epochs again with resources from the accountant.

The db pool's only task should be storing stuff imo

services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,905 @@
...
+/// Aggregates recorded RP request counts per epoch and submits them as billing votes.
+///
+/// Constructed via [`OprfAccountantService::new`]; driven by repeatedly calling
+/// [`OprfAccountantService::tick`] (typically via [`OprfAccountantService::run`]).
+#[derive(Clone)]
+#[expect(
+    dead_code,
+    reason = "unused fields are used in the future implementation of vote submission"
+)]
+pub struct OprfAccountantService {
```
0xThemis: nit: Everything in this file is pub. Is anything needed outside of this crate?

services/oprf-accountant/src/api.rs
```diff
@@ -39,13 +40,21 @@ impl From<&NullifierOprfRequestAuthV1> for BillableRpRequest {
     }
 }
 
+#[instrument(level = "trace", skip_all)]
```
0xThemis: should be info imo. Also I think we should add a Uuid to the worker at the node to link traces afterwards. 

Something to follow up on

services/oprf-accountant/src/config.rs
```diff
@@ -12,4 +16,43 @@ pub struct OprfAccountantConfig {
...
+        default = "OprfAccountantConfig::default_submit_interval",
+        with = "humantime_serde"
+    )]
+    pub submit_interval: Duration,
+
+    /// A additional offset to be added to the start of the voting window.
+    ///
+    /// This is used to ensure that the OPRF nodes have enough time to send their requests
+    /// to the accountant before the voting window starts.
+    /// The offset should be ~2x the flush interval of the `AccountantBatcher`.
```
0xThemis: "..should be at least.."

I think in praxis it will be a lot higher then two 2x

services/oprf-accountant/src/config.rs
```diff
@@ -12,4 +16,43 @@ pub struct OprfAccountantConfig {
...
+
+    /// The `BillingContract` submit interval.
+    ///
+    /// Must be at most half the smallest `votingWindow` (and `epochLength`) the contract is
+    /// configured with, otherwise the accountant may miss an epoch's voting window entirely.
+    #[serde(
+        default = "OprfAccountantConfig::default_submit_interval",
+        with = "humantime_serde"
+    )]
+    pub submit_interval: Duration,
```
0xThemis: `submit_interval` is misleading no?

This is the tick interval right? Can be misleading as we don't necessarily submit.

services/oprf-accountant/src/lib.rs
```diff
@@ -21,24 +21,85 @@
...
 pub mod api;
 pub mod config;
 pub mod metrics;
 pub mod postgres;
+pub mod timing_watcher;
 
 #[derive(Clone)]
-struct AppState(PostgresDb);
+struct AppState {
+    db: PostgresDb,
```
0xThemis: do we need the DB in the AppState?

services/oprf-accountant/src/lib.rs
```diff
@@ -21,24 +21,85 @@
...
-    Router::new().merge(api::routes()).with_state(AppState(db))
+pub async fn start(
+    config: &OprfAccountantConfig,
+    db: PostgresDb,
+    cancellation_token: CancellationToken,
+) -> eyre::Result<Router> {
+    let provider =
+        taceo_nodes_common::web3::HttpRpcProviderBuilder::with_config(&config.rpc_provider_config)
+            .build()?
+            .inner();
```
0xThemis, line 61: remove the inner and use the `HttpRpcProvider` to have type safety in the methods

services/oprf-accountant/src/postgres.rs
```diff
@@ -52,16 +55,18 @@ impl PostgresDb {
...
         let rp_ids = rp_requests
             .iter()
             .map(|r| r.rp_id.into_inner() as i64)
             .collect_vec();
-        // let epochs: Vec<i64> = rp_requests.iter().map(|r| r.epoch).collect();
-        // TODO get epochs
-        let epochs: Vec<i64> = Vec::new();
+        let epochs = rp_requests
+            .iter()
+            .map(|r| i64::from(epoch_for_timestamp(timing_eras, r.expires_at)))
```
0xThemis: as already said, feels weird that the DB is doing that. Shouldn't be the responsibility of the DB imo

services/oprf-accountant/src/timing_watcher.rs
```diff
@@ -0,0 +1,177 @@
...
+/// Spawns a background task that keeps a [`TimingEras`] snapshot in sync with the
+/// `BillingContract`'s `TimingUpdated` events.
+///
+/// [`TimingWatcher`] itself holds no state: [`TimingWatcher::init`] returns the shared
+/// [`TimingEras`] handle that the spawned task updates in place.
+#[derive(Clone)]
+pub(crate) struct TimingWatcher;
+
+impl TimingWatcher {
+    /// Fetches the current era history via `http_provider`, then spawns a background task that
```
0xThemis: Reverse the comment here.
Subscribe then fetch becuase that is what it does and otherwise it would be broken anyways

services/oprf-accountant/src/timing_watcher.rs
```diff
@@ -0,0 +1,177 @@
...
+        // Fetches the initial era history after we subscribe, so we don't miss any eras that were appended between the initial fetch and the subscription.
+        *timing_eras.lock().expect("not poisoned") = http_contract
+            .getEras()
+            .call()
+            .await
+            .context("while fetching initial timing eras")?;
+
+        // Appends each decoded `TimingUpdated` event to `timing_eras` as it arrives, until
+        // cancelled or the event stream ends.
+        tokio::spawn({
```
0xThemis: I think this task should return an eyre::Result and be propagated to call site. 

Then during shutdown we can await that task and log the error (like for key-gen and the nodes). Otherwise we get a ExitCode::OK eventhough this stream ended with an error and that was the reason we failed.

services/oprf-accountant/src/timing_watcher.rs
```diff
@@ -0,0 +1,177 @@
...
+                                    };
+                                    tracing::trace!(?era, "received TimingUpdated event; appending to timing eras");
+                                    let mut timing_eras = timing_eras.lock().expect("not poisoned");
+                                    // we could get a event for a new TimingEra that we already have in our list if the event happened right after we fetch them initially or when we fetch them all in try_reconnect.
+                                    if timing_eras.last().is_some_and(|last| last.startEpoch >= era.startEpoch) {
+                                        tracing::warn!(
+                                            "received TimingUpdated event with non-increasing startEpoch; skipping"
+                                        );
+                                    } else {
+                                        timing_eras.push(era);
```
0xThemis: should we just call .getEras() and fetch  everything? Otherwise there is a race condition with the initial HTTP request

services/oprf-accountant/src/lib.rs
fabian1409: later for the get route for RPs to get the requests they paid for
but depending on the auth for that it may make more sense for the `OprfAccountantService` to handle it.

but let remove it for now

services/oprf-accountant/src/lib.rs
fabian1409: then I cant use the mock transport provider in tests

services/oprf-accountant/src/accountant_service.rs
```diff
@@ -0,0 +1,936 @@
...
+                        epoch,
+                        epoch_cursor,
+                        "epoch cursor is more than one behind; some epochs may have been skipped"
+                    );
+                }
+                if epoch as i64 > epoch_cursor {
+                    tracing::info!(epoch, ?counts, "submitting billing vote for epoch");
+                    self.vote_for_epoch(epoch, counts).await?;
+                    tracing::info!(epoch, "successfully submitted billing vote for epoch");
+                    self.db.set_epoch_cursor(epoch).await?;
```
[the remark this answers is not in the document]
fabian1409: this will be added in a followup PR

fabian1409:
replaced by #858

PR #855, opened 2026-07-10
feat: allow binding uniqueness proofs to an existing session

kilianglas:
## Summary

Allows a Uniqueness Proof request to carry an existing `session_id`, cryptographically binding the proof to that session: the proof carries `SessionId.commitment` as its `id_commitment` public signal, proving in-circuit that the session and the nullifier belong to the same World ID. Requires no circuit changes, the `id_commitment` check in `oprf_nullifier.circom` is proof-type-independent and was previously always disabled (`0`) for uniqueness proofs.

## New flow for a bound session id

1. RP sends a `ProofType::CreateSession` request. 
2. Authenticator generates the `session_id` and responds with corresponding session proofs.
3. RP sends a `ProofType::Uniqueness` request that now also carries `Some(session_id)`
4. Authenticator generates a uniqueness proof as before, but now also passes `session_id` as a public input.

Follow-up PRs may streamline this flow s.t. the initial request is not required, but the `session_id` can be generated as part of the uniqueness proof instead.

## Changes:

- `world-id-primitives`: `validate_proof_type()` now allows `session_id` on `ProofType::Uniqueness`; new `ProofRequest::binds_session()`; `validate_response()` requires bound responses to echo the `session_id` (and still rejects a `session_id` on plain uniqueness responses).
- `world-id-authenticator`: `generate_proof()` resolves `(session_id, r)` for binding requests from the caller-cached `session_id_r_seed`, validating it opens the session's commitment. A missing seed fails with the new `SessionSeedRequired` error.
Note: Re-deriving `r` requires a session-type request, since the uniqueness request's RP signature cannot authorize the seed query. This friction may be removed in a follow-up PR.
- Docs: short spec section on session binding, including the verifier obligation (a proof verified with the signal at `0` is valid but unbound) and the intentional-linkability caveat.

On-chain, bound proofs can already be verified via `verifyProofAndSignals` by passing the session's commitment as the `sessionId` signal; the convenience `verify()` entry point pins the signal to `0` and rejects bound proofs. A dedicated entry point for bound proofs may land in a follow-up PR.


dkales (dismissed):

paolodamico:
we should update the session proofs flow doc (https://github.com/worldcoin/world-id-protocol/tree/main/docs/world-id-4-specs#session-proofs) from this

paolodamico (commented):
1. _updating docs_
2. conceptual: I wonder if it should be generally allowed to request a session_id without the uniqueness proof. it could be a footgun for RPs because then RPs would have to immediately request a session proof to verify it's valid.
3. let's document details on privacy implications, for example, you could always request another uniqueness proof with the same `session_id`. at the very least, when doing this, the authenticator's UI MUST NOT display any "anonymous" / "one-time" / etc language.

kilianglas:
> we should update the session proofs flow doc (https://github.com/worldcoin/world-id-protocol/tree/main/docs/world-id-4-specs#session-proofs) from this

I'd wait until I've implemented the final version of this (single uniqueness proof flow).

kilianglas:
> conceptual: I wonder if it should be generally allowed to request a session_id without the uniqueness proof. it could be a footgun for RPs because then RPs would have to immediately request a session proof to verify it's valid.

I'd say yes. Session proofs might be useful without a uniqueness proof, but might be wrong here.

kilianglas:
> et's document details on privacy implications, for example, you could always request another uniqueness proof with the same session_id. at the very least, when doing this, the authenticator's UI MUST NOT display any "anonymous" / "one-time" / etc language.

I don't fully get the privacy implications here. What's the point of requesting a different uniqunesess proof with the same session id?

0xOsiris (dismissed):
This is functionally identical to passing the `sessionId` as the signal to the uniqueness proof correct?

kilianglas:
> This is functionally identical to passing the sessionId as the signal to the uniqueness proof correct?

No, it is not. This also verifier, that the `session_id` and `nullifier` were created by the same world id, by essentially doing a session proof and a nullifier/ uniquneness proof in a single proof.

crates/authenticator/src/prove.rs
```diff
@@ -527,6 +532,19 @@ impl Authenticator {
 
         Ok(generate_ownership_proof_with_prover(input, prover)?)
     }
+
+    fn validate_cached_session_r_seed(
```
Dzejkop, line 536: nit: but maybe just `validate_session_r_seed`

`cached` implies there might be caching happening in this method

Dzejkop (dismissed):

crates/authenticator/src/prove.rs
```diff
@@ -309,7 +308,15 @@ impl Authenticator {
 
         // 2. Resolve session seed
         let (resolved_session_id, resolved_session_seed) = match proof_request.proof_type {
-            ProofType::Uniqueness => (None, None),
+            ProofType::Uniqueness => match proof_request.session_id {
+                // Bind the proof to the existing session. Requires the cached `r`.
+                Some(session_id) => {
+                    let seed = session_id_r_seed.ok_or(AuthenticatorError::SessionSeedRequired)?;
```
paolodamico, line 314: this can be tackled in a separate PR, but I don't think we should fail here. if the RP provides a valid session_id we should recompute, especially because the RP has no control over what the user caches

crates/authenticator/src/prove.rs
paolodamico: this function also looks suspiciously close to functionality elsewhere, I recommend moving this to SessionId, it seems potentially problematic to have it here, if we're implementing validation here it means we could miss it elsewhere.

crates/core/tests/generate_proof.rs
```diff
@@ -363,6 +365,90 @@ async fn e2e_authenticator_generate_proof() -> Result<()> {
...
+            nullifier_for_binding.clone(),
+            &credentials,
+            None,
+            None,
+        )
+        .await
+        .unwrap_err();
+    assert!(matches!(err, AuthenticatorError::SessionSeedRequired));
+
+    // a seed that does not open the session's commitment is rejected
```
paolodamico:
```suggestion
    // a seed that does not match the session's commitment is rejected
```

paolodamico (dismissed):
**high**. if I think about the problem fresh, the conceptual solution in my mind is that RPs use a uniqueness proof to gate access to creating a session, so uniqueness proof -> session, while the other way around may be more confusing (though correct). this also maintains the clear separation that a uniqueness proof is anonymous one-time use and avoids authenticators having to manage this variant. for consideration @kilianglas

crates/authenticator/src/prove.rs
kilianglas: Will follow up on this.

dkales (approved):

docs/world-id-4-specs/README.md
```diff
@@ -279,6 +279,13 @@ rp->>rp: verify proof (checking sessionId == C' in verifier contract)
 - The raison d'être is simply to allow usage of the same ZK circuit as for Uniqueness Proofs. Reducing the number of circuits is currently a priority because of the size of the circuits needed to be bundled in Authenticator clients. As World ID moves to a different proving system, this type will no longer be required.
 - Session Proofs use a randomized `action` as circuit input. This randomized `action` ensures the circuit's nullifier output is unique per proof, preserving the one-time use property. It is verified internally within the circuit. It does not affect `r` derivation.
 
+**Binding Uniqueness Proofs to a Session**
+
+- A Uniqueness Proof request may include an existing `sessionId`. The proof then carries the session's commitment `C` as its `id_commitment` public signal, proving in-circuit that the session and the nullifier belong to the same World ID.
```
paolodamico, line 284: This comment is outdated. It cannot include an existing session id.

docs/world-id-4-specs/README.md
```diff
@@ -279,6 +279,13 @@ rp->>rp: verify proof (checking sessionId == C' in verifier contract)
...
+- The Authenticator requires the cached `r` for this; re-deriving `r` is only possible through a session-type request.
+- Verifiers MUST check the proof against the session's commitment — with the signal at `0` the proof is valid but unbound. On-chain, `verifyProofAndSignals` accepts the commitment as the `sessionId` signal; the convenience `verify()` entry point pins the signal to `0` and rejects bound proofs.
+- Binding one `sessionId` to Uniqueness Proofs under different actions intentionally links those actions to the same World ID; Authenticators MUST clearly surface this to users.
```
paolodamico, line 287: Also outdated.

paolodamico (approved):
Approving because outstanding items can be tackled in the stacked PR

PR #862, opened 2026-07-13
feat: add verifyWithSession verifier entry point for session-bound proofs

kilianglas:
Adds a verifier entry point for Uniqueness Proofs bound to a session commitment. The client side (#855) emits proofs whose `session_id` public signal carries the commitment; `verify()` pins that signal to `0`, so bound proofs need their own entry point. Verification logic is unchanged — `verifyProofAndSignals` already takes `sessionId`.

Extends the V3 verifier added in #882. No V2 file is touched.

## Contracts

- `UnreleasedIWorldIDVerifierV3` declares `verifyWithSession(...)`.
- `UnreleasedWorldIDVerifierV3` implements it: requires the uniqueness action prefix (MSB `0x00`, `InvalidAction`) and `sessionId != 0` (`InvalidSessionId`), then delegates to `verifyProofAndSignals`.
- `verify()` and `verifySession()` are unchanged; bound proofs are rejected by `verify()` and unbound proofs by `verifyWithSession`, so binding is explicit in both directions.

## Rust

- `RegistryTestContext` deploys `WorldIDVerifierV3` behind the proxy.
- `generate-solidity-fixtures` gains a bound-proof section: binds the request to the session fixture's id (cached `r`), asserts the nullifier is unchanged, and verifies on-chain.
- The e2e positively verifies a bound proof via `verifyWithSession`, alongside the existing `verify()`-rejects assertion.

## Tests & fixtures

- `WorldIDVerifierV3Test` goes from 4 to 11 tests: happy path, bound proof rejected by `verify()`, unbound rejected by `verifyWithSession`, wrong session id, action-prefix checks, zero session id.
- Fixtures in `WorldIDVerifierTest.t.sol` (shared mocks and merkle root) and the V3 test are regenerated as one consistent set — they can only be replaced together. The session fixture now carries its own `sessionNonce` and a `0x02`-prefixed session action.


contracts/src/core/interfaces/IWorldIDVerifierV2.sol
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @title IWorldIDVerifierV2
+ * @author World Contributors
+ * @notice Interface for verifying World ID proofs (Uniqueness and Session proofs).
+ * @dev V2 enforces the action-prefix convention on the convenience entry points (`verify`
+ *  requires the action's most significant byte to be `0x00`, `verifySession` requires `0x02`)
+ *  and adds `verifyWithSession` for Uniqueness Proofs bound to an existing session.
+ */
+interface IWorldIDVerifierV2 is IWorldIDVerifier {
```
Dzejkop, line 13: If we're not including it in V2 imo we should do a V3 here. Since WorldIDVerifierV2 already exists - but there were not interface changes

contracts/test/core/WorldIDVerifierTest.t.sol
```diff
@@ -13,7 +13,7 @@ import {ICredentialSchemaIssuerRegistry} from "../../src/core/interfaces/ICreden
 uint64 constant credentialIssuerIdCorrect = 1;
 uint64 constant credentialIssuerIdWrong = 2;
 
-uint64 constant rpIdCorrect = 0x1a6ccf8f70e5de68;
+uint64 constant rpIdCorrect = 0x387df34f862cd4e;
```
Dzejkop, line 16: ?

contracts/src/core/WorldIDVerifierV2.sol
```diff
@@ -3,23 +3,17 @@ pragma solidity ^0.8.13;
 
 import {WorldIDVerifier} from "./WorldIDVerifier.sol";
 import {IWorldIDVerifier} from "./interfaces/IWorldIDVerifier.sol";
+import {IWorldIDVerifierV2} from "./interfaces/IWorldIDVerifierV2.sol";
 
 /**
- * @title WorldIDVerifier
+ * @title WorldIDVerifierV2
```
paolodamico: should we still update V2 or is this V3 now? CC @Dzejkop

contracts/test/core/WorldIDVerifierTest.t.sol
kilianglas: This is generated by the fixture generator and we needed to re-generate for the new bound uniqueness proof fixture.

paolodamico, line 16: given these tests were part of an audit, is it possible to introduce a new module for V3 tests?

paolodamico (approved):

PR #863, opened 2026-07-13
refactor: add missing IWorldIDVerifierV2 interface

kilianglas:
## Summary

`WorldIDVerifierV2` is the only versioned core contract without a matching versioned interface — `WorldIDRegistryV2` pairs with `IWorldIDRegistryV2 is IWorldIDRegistry`, while the verifier V2 declares its `InvalidAction` error directly on the contract. This adds the missing `IWorldIDVerifierV2` so the verifier follows the same pattern, and so future verifier versions (e.g. a V3 adding new entry points) have a proper interface chain to extend.

- New `IWorldIDVerifierV2 is IWorldIDVerifier`: declares `InvalidAction` and documents V2's action-prefix semantics (`verify` requires MSB `0x00`, `verifySession` requires `0x02`).
- `WorldIDVerifierV2 is IWorldIDVerifierV2, WorldIDVerifier` (mirrors `WorldIDRegistryV2`): local error declaration removed; overrides now name both bases (`override(IWorldIDVerifier, WorldIDVerifier)`), as required by the diamond inheritance and consistent with the registry.
- Tests reference the error via the interface (`IWorldIDVerifierV2.InvalidAction.selector`), matching the registry test idiom.
- Also fixes the stale `@title WorldIDVerifier` on the V2 contract.

No behavior change: same functions, same checks, same error selector.

## Test plan

- [ ] `cd contracts && forge test` — 331 tests green

### DIFF
--- a/contracts/src/core/WorldIDVerifierV2.sol
+++ b/contracts/src/core/WorldIDVerifierV2.sol
@@ -3,23 +3,17 @@ pragma solidity ^0.8.13;
 
 import {WorldIDVerifier} from "./WorldIDVerifier.sol";
 import {IWorldIDVerifier} from "./interfaces/IWorldIDVerifier.sol";
+import {IWorldIDVerifierV2} from "./interfaces/IWorldIDVerifierV2.sol";
 
 /**
- * @title WorldIDVerifier
+ * @title WorldIDVerifierV2
  * @author World Contributors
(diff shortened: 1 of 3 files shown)


kilianglas:
Superseded: the V2 interface will be introduced together with verifyWithSession in #862 instead.

PR #864, opened 2026-07-14
feat: add currentVoteEpoch oprf-accountant helper to BillingContract

fabian1409:
> [!NOTE]
> **Low Risk**
> Additive view-only API with no storage or authorization changes; timing/era edge cases are covered by extensive new tests, though incorrect epoch selection would mislead off-chain voters.
> 
> **Overview**
> Adds **`currentVoteEpoch(signer)`** on `BillingContract` / `IBillingContract`: a single read that returns the latest **fully elapsed** epoch (not the in-flight one), its `[epochStart, epochEnd)` span, `votingWindowEnd`, whether that signer has **closed** a chunked vote (`hasVoted`), and `block.timestamp`. Intended so an OPRF billing accountant can decide when/what to vote without extra RPCs.
> 
> The era walk mirrors `epochWatermarks` / `_latestClosedEpoch` logic but keys off each era’s span start rather than voting-window close, so mid-epoch callers still see the previous elapsed epoch. `exists` is false only while epoch 0 is still running; `alreadyVoted` is per-signer and resets after `finalizeEpochs` prunes transient submitter state.
> 
> Foundry coverage adds a dedicated `currentVoteEpoch` test block (genesis, progression, per-signer votes, incomplete chunks, finalization, timing-era changes, `EpochTooLarge`, and multi-epoch participation).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c754d3b49ecb67d031b913d40cd75c82167035c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/contracts/src/core/BillingContract.sol
+++ b/contracts/src/core/BillingContract.sol
@@ -468,6 +468,52 @@ contract BillingContract is WorldIDBase, IBillingContract {
         return _timingEras;
     }
 
+    /// @inheritdoc IBillingContract
+    function currentVoteEpoch(address signer)
+        external
+        view
+        virtual
+        onlyProxy
+        onlyInitialized
+        returns (
+            bool exists,
+            uint32 epoch,
+            uint64 epochStart,
+            uint64 epochEnd,
+            uint64 votingWindowEnd,
+            bool alreadyVoted,
+            uint64 blockTime
+        )
+    {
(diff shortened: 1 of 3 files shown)


fabian1409:
approach to billing will be different, so this is obsolete

PR #866, opened 2026-07-14
refactor!: collapse create_session proof type into session x create

kilianglas:
Removes `ProofType::CreateSession` and redesigns the request model: `proof_type` shrinks to `{uniqueness, session}` and `session_id` becomes a three-state field. Creating a session was always create-and-prove (a session proof in the same response), so it collapses into `proof_type: "session"` + `session_id: "create"` with no capability change. Hence, there are only two proof types now, Uniqueness Proofs and Session Proofs. Both can create a session, only session proofs can be bound to an existing session. Therefore, sessions can either be bound to a nullifier, or can be started without any binding. Note, that Uniquess Proofs can still be requested without creating or binding a session.

Stacked on #862.

## Request model

New `SessionRef` type carried by `ProofRequest.session_id`:

| wire value | state |
|---|---|
| absent / `null` | `SessionRef::None` |
| `"create"` | `SessionRef::Create` — mint a fresh session, prove it in the same response |
| `"session_<hex>"` | `SessionRef::Existing(SessionId)` |

Note: The `SessionRef` is **not** in the signed part of the request.

Validation matrix (`validate_proof_type`):

| | `null` | `"create"` | `"session_<hex>"` |
|---|---|---|---|
| `uniqueness` | ✅ plain | ✅ create bound session| ❌ |
| `session` | ❌ | ✅ (was `create_session`) | ✅ |

`action` remains forbidden for `proof_type: "session"` (both sub-states).

Different combinations have different use cases:
* `uniqueness`, `null`: Standalone uniqueness proof that does not require session. 
* `uniqueness`, `"create"`: Uniqueness proof that requires bound session for later session proof.
* `session`, `"create"`: Create new standalone session to prove human continuity without uniqueness guarantees. Used e.g., in the upcoming DeepFace proof type.
* `session`, `"session_<hex>"`: Session proof for existing session. This is used for any `session_id`, uniqueness bound or not. 

## Changes to OPRF nodes

To make this work, the auth module of the ORPF nodes had to be slightly changed. OPRF queries with prefix `0x01` (used to obtain the session seed), can now originate from Uniqueness Proof RP requests. Those have an `action`  field iin the signed data. Hence, in order to verify the RP Request signature, the OPRF request needs to carry this `action`, which differs from `action = oprf_seed` which is used for the query. We introduce a new field called `rp_signature_verification` on `NullifierOprfRequestAuthV1` struct. 

## WIP-101 RPs
Important: Combined uniqueness + create session requests are not yet supported. This will be implemented in a follow-up PR.


crates/primitives/src/session.rs
```diff
@@ -245,6 +245,137 @@ impl<'de> Deserialize<'de> for SessionId {
...
+                )
+            }
+
+            fn visit_none<E: serde::de::Error>(self) -> Result<Self::Value, E> {
+                Ok(SessionRef::None)
+            }
+
+            fn visit_unit<E: serde::de::Error>(self) -> Result<Self::Value, E> {
+                Ok(SessionRef::None)
+            }
```
dkales, line 341: Is visit unit needed? The serializer does not actually use it?

crates/authenticator/src/prove.rs
```diff
@@ -208,20 +208,23 @@ impl Authenticator {
...
 
         let mut rng = rand::rngs::OsRng;
 
         let oprf_seed = match proof_request.session_id {
-            Some(session_id) => session_id.oprf_seed,
-            None => SessionId::generate_oprf_seed(&mut rng),
+            SessionRef::Existing(session_id) => session_id.oprf_seed,
+            SessionRef::Create => SessionId::generate_oprf_seed(&mut rng),
+            SessionRef::None => {
+                unreachable!("SessionRef::None should be handled by the guard above")
```
dkales: why not move the above if block in here then?

crates/primitives/src/request/mod.rs
```diff
@@ -104,12 +100,14 @@ pub struct ProofRequest {
...
-    /// Required for [`ProofType::Session`], forbidden for [`ProofType::CreateSession`],
-    /// optional for [`ProofType::Uniqueness`] to bind the proof to an existing session
-    /// (see [`Self::binds_session`]).
+    /// Three states: absent/`null` (no session), `"create"` (mint a fresh session),
+    /// or an existing `"session_"`-prefixed id. For [`ProofType::Uniqueness`], all
+    /// three are valid; for [`ProofType::Session`], `"create"` or an existing id is
+    /// required (see [`Self::binds_session`]).
     /// The proof will only be valid if the session ID is meant for this context and
     /// this particular World ID holder.
-    pub session_id: Option<SessionId>,
```
dkales, line 112: An old request might actually parse as a SessionRef::Existing here, depending on how the autogenerated serde derive handles the option here. I guess this is fine though and maybe even wanted?

crates/primitives/src/request/mod.rs
dkales, line 100: Old `create_session` requests will now fail for sure though.

0xThemis (approved):

dkales (approved):

crates/primitives/src/request/mod.rs
kilianglas: Yes. @paolodamico is this an issue?

crates/primitives/src/request/mod.rs
kilianglas: I think this is fine. We accept:
* old `"proof_type": "uniqneness"` without a `session_id` field 
* old `"proof_type": "session"` with an existing `session_id` field
What is we do not accept is `"proof_type": "create_session"` since this no longer exists. To my understanding this is not used by anyone right now, so this should not be an issue. Thoughst @paolodamico @Dzejkop ?

crates/primitives/src/session.rs
kilianglas: I think visit_unit should be implemented as best practice because some Deserlizers might not call deserliaze_option, but just fallback to deserialize_any and then land on visit_unit. This is a very unlikely edge case, though.

0xThemis (dismissed):

dkales (dismissed):

0xThemis (dismissed):

0xForerunner:
@kilianglas Why do we disallow `uniqueness + session_<hex>`?

0xForerunner:
It's worth noting that this PR allows oprf nodes to link the initial uniqueness action and the session seed. I don't really think that's a problem in practice, just wanted to flag it.

crates/proof/src/oprf_query.rs
```diff
@@ -301,6 +305,13 @@ impl<'a> OprfEntrypoint<'a> {
             rng,
         )?;
 
+        let rp_signature_verification = match (proof_request.proof_type, proof_request.action) {
+            (ProofType::Uniqueness, Some(action)) => {
+                Some(RpSignatureVerification::UniquenessAction { action })
```
0xForerunner, line 310: The signed message doesn't include proof_type or session_id, which means these fields can be changed and the signature is still valid. We probably want to fix this.

0xForerunner (changes requested):

kilianglas:
> @kilianglas Why do we disallow `uniqueness + session_<hex>`?

We thought it might be an anonymity footgun if we allow RPs to bind uniqueness proofs to existing sessions.

kilianglas:
> It's worth noting that this PR allows oprf nodes to link the initial uniqueness action and the session seed. I don't really think that's a problem in practice, just wanted to flag it.

Yes, that's true. Didn't think this was an issue, especially since `r` is cached anyway by the authenticator, session proofs don't need the `oprf_seed` round trip in most cases. Can you think of any scenario where this might be an issue?

crates/proof/src/oprf_query.rs
kilianglas: This is a valid more general question. The proof type is implicitly encoded in the action (action with `0x00` prefix is present -> uniqueness, no action -> session). The session id field is currently not signed over by the RP, and there might be ways to abuse this (need to think about this). Since this is pre-existing behavior, though, I don't think this should be addressed in this PR.

0xForerunner (dismissed):

dkales (approved):

Dzejkop (approved):

PR #880, opened 2026-08-04
feat: build authenticator-attestation on nargo beta.11

kilianglas:
Moves `crates/proof/noir/authenticator-attestation` from nargo `1.0.0-beta.20` to `1.0.0-beta.11`, so every Noir package in the repo builds with a single toolchain.

## Why

The published `provekit-*` crates depend on the republished `provekit_acir` / `provekit_noirc_abi` forks, which exist only at `1.0.0-beta.11-alpha.x`. `world-id-proof` is `publish = true`, so it cannot take git dependencies on provekit `main` (which does target beta.20). A package that only builds on a newer nargo therefore cannot be proven, only tested.

beta.20 was not a requirement of the circuit itself. It arrived through `noir_ecdsa v0.4.x`, which pins `bigcurve v0.14.0-1`; that generates its curve impls with the `#[derive_curve_impl]` comptime macro, which beta.11 cannot expand. `noir_ecdsa v0.2.9` pins `bigcurve v0.11.0-1` and builds on beta.11 with the same ECDSA implementation and identical low-S semantics.

## Changes

- Pin the beta.11 dependency generation: `noir_ecdsa v0.2.9`, `bigcurve v0.11.0-1`, `bignum v0.8.0-2`, `poseidon2 v0.5.0-beta.0` (`perm::x5_8` rather than `permutation::t8`).
- Replace the vendored `src/eddsa.nr` with TACEO's upstream `eddsa_poseidon2` from `oprf-service` at `taceo-oprf-v0.12.0`. It sits alongside `babyjubjub` in the same tree and the two are a matched pair, so they are pinned together. The upstream file is byte-identical to the tree Least Authority audited in January 2026.
- Assert `r != 0` in `crates/proof/noir/authenticator-attestation/src/aat.nr`. `noir_ecdsa` only gained this check in v0.4.1; without it an `r = 0` signature verifies against any public key, since the verification equation collapses to `R = u1*G` and drops the public key. Covered by a new `test_aat_zero_r_rejected`.
- Point the `authenticator-attestation` CI job at beta.11, matching `ownership-proof`.
- Document the toolchain constraint, the pin rationale, and the two pin-change hazards in `crates/proof/noir/README.md`.


.github/workflows/noir-ci.yml
```diff
@@ -34,8 +34,8 @@ jobs:
       - uses: actions/checkout@v6
       - uses: noir-lang/noirup@7dbe69ccc78877f0200ffa5a40836c953d2cfd8f # v0.1.4
         with:
-          # This package requires beta.20: the pure-Noir ECDSA stack and the
-          # oprf-nr babyjubjub library target it (matching provekit git main).
-          # See crates/proof/noir/README.md.
-          toolchain: v1.0.0-beta.20
+          # Same toolchain as ownership-proof: the published provekit crates
```
paolodamico: I'd drop this comment, at this point is self-evident the same noir toolchain is used across the board

crates/proof/noir/authenticator-attestation/src/aat.nr
```diff
@@ -57,6 +64,14 @@ pub fn verify_aat(aat: AuthenticatorAssertionToken) {
     let public_key_y = Secp256r1_Fq::from_be_bytes(aat.takt.assertion_key_y);
     let r = signature_scalar(aat.signature, 0);
     let s = signature_scalar(aat.signature, 32);
+
+    // `noir_ecdsa` only rejects `r == 0` from v0.4.1 onwards, and we pin v0.2.9
+    // (see crates/proof/noir/README.md), so the check lives here instead. With
+    // `r = 0` the verification equation collapses to `R = u1*G`, dropping the
+    // public key entirely: a single forgery would then verify under *every*
+    // key. `s == 0` is already rejected by the `s * s^-1 == 1` constraint.
+    assert(r != Secp256r1_Fr::zero(), "AAT signature r value is zero");
```
paolodamico, line 73: good catch!

crates/proof/noir/authenticator-attestation/src/utils.nr
```diff
@@ -14,3 +14,18 @@ pub(crate) fn p256_be_coord_to_field(coord: [u8; 32]) -> (Field, Field) {
...
+    let bytes: [u8; 4] = value.to_be_bytes();
+    assert((bytes[0] != 0) | (bytes[1] != 0), "value must require a 4-byte CBOR uint encoding");
+}
+
+/// Asserts CWT `exp` freshness: `now < exp` (expired at `exp`, RFC 8392).
+///
+/// `now` MUST be a verifier-supplied public input; a prover-chosen private
+/// timestamp would make the check vacuous.
+pub(crate) fn assert_not_expired(exp: Field, now: Field) {
+    std::field::bn254::assert_lt(now, exp);
```
paolodamico, line 30: @dkales afaik we don't hardcode field assumptions anywhere else, I'm not sure this is correct

paolodamico (approved):

crates/proof/noir/authenticator-attestation/src/utils.nr
dkales: There are some assumptions in the circuits on the concrete field (like the poseidon instance for example, or the fact that BJJ curve is base-field compatible to it). But < should also just work here.

kilianglas: We are already using `poseidon2::bn254`, so that would break first if the backend I guess.

PR #882, opened 2026-08-04
fix: reject zero sessionId in verifySession

kilianglas:
`verifySession` accepts `sessionId == 0`, which is the circuit's "no session" sentinel and is satisfiable by any World ID. A session record left at zero can therefore be satisfied by an arbitrary World ID, voiding the "same user + RP pair across requests" continuity guarantee that `verifySession` is meant to provide.

Adds `WorldIDVerifierV3`, which overrides `verifySession` to reject a zero `sessionId`. Purely additive — three new files, no existing file is modified.

Reported via HackerOne #3912490.

### Why the sentinel is universally satisfiable

`circom/client_side_proofs/oprf_nullifier.circom:183` constrains the commitment as a disjunction:

```
id_commitment * (id_commitment - computed_id_commitment) === 0;
```

It is satisfied either when `id_commitment == H(DS_C, mt_index, id_commitment_r)` or when `id_commitment == 0`. The zero branch places no constraint on `mt_index`, so it is not merely "unbound" — it is provable by any identity in the tree. Every Uniqueness Proof uses that branch legitimately, so the sentinel cannot be removed from the circuit; the verifier has to reject it at the boundary instead.


contracts/test/core/WorldIDVerifierV3Test.t.sol
```diff
@@ -0,0 +1,123 @@
...
+            oprfKeyRegistry,
+            groth16Verifier,
+            minExpirationThreshold
+        );
+        ERC1967Proxy proxy = new ERC1967Proxy(address(implementation), initData);
+        verifier = WorldIDVerifierV3(address(proxy));
+        verifier.updateOprfKeyRegistry(oprfKeyRegistry);
+    }
+
+    function test_SessionRevertsWhenSessionIdZero() public {
```
Dzejkop: redundant with respect to the fuzz test below?

contracts/src/core/UnreleasedWorldIDVerifierV3.sol
```diff
@@ -0,0 +1,52 @@
...
+            action,
+            rpId,
+            nonce,
+            signalHash,
+            expiresAtMin,
+            issuerSchemaId,
+            credentialGenesisIssuedAtMin,
+            sessionId,
+            zeroKnowledgeProof
+        );
```
Dzejkop, line 50: could leave just 

```
if (sessionId == 0) {
    revert InvalidSessionId();
}
```

and then call the super method?
so something like:

```suggestion
        if (sessionId == 0) {
            revert InvalidSessionId();
        }

        super.verifySession(rpId, ...)
```

kilianglas: I think this doesn't compile since super.verifySession is external. this.... also doesn't work.

contracts/src/core/UnreleasedWorldIDVerifierV3.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
```
paolodamico, line 1: Can we please append the Unreleased prefix to this file's name to avoid confusion? This can be removed once it's shipped.

paolodamico (dismissed):

Dzejkop (approved):

paolodamico (approved):

PR #884, opened 2026-08-04
feat: decouple the TAKT from the AAT and gate verification behind one entrypoint

kilianglas:
Aligns the WIP-106 attestation library with the current spec and makes it consumable from another Noir package.

## The TAKT is fully decoupled from the AAT

- The circuit no longer reserializes the TAKT: `serialize_takt`, `compress_bjj_point`, `FIELD_HALF` and its serialization KAT are deleted. **50742 → 49113 ACIR (−3.2%)**. 
- The Rust `AuthenticatorAssertionToken::new` now validates AAT claims independently; it no longer parses a TAKT or retains its attested key. Pairing and key-consistency checks belong to the carrying/verifying layer.
- `-80000` renamed `signal` → `cdh`.
- New `aat_known_answer_sig_structure_and_signature` KAT. The TAKT had one, the AAT did not, which is how the Noir fixture could silently drift from the Rust encoder. It caught a bug on first run: a 159-byte payload takes the 1-byte CBOR length, not the 2-byte form the old 417-byte payload needed.

## Lifetime caps and fixed-width claims

- `verify_aat` / `verify_takt` enforce the WIP-106 maximum remaining lifetimes: 30 min (AAT) and 7d (TAKT), inclusive. Minting deliberately does not enforce them.
- `aud` is now a fixed 8-byte big-endian byte string, so the full `RpId` u64 range encodes at constant width (ids as small as `1` are in use). Payload is now 163 bytes, `Sig_structure` 182.
- Rust `exp` is `u32`, validated to `[2^16, 2^32)` on both tokens.
- A new test pins noir-ecdsa's low-S rejection so a version bump cannot silently drop it.

## One public entrypoint

```rust
verify_takt(...)   // pub(crate)
verify_aat(...)    // pub(crate)
verify_attestation(aat, takt, trust_anchor_key_x, _y, now)   // pub, at the crate root
```

The composite feeds `verify_aat` the key read from the TAKT it just verified, so the binding holds by construction.
Struct fields are now `pub`: they were module-private, so no external package could construct either token.

## Rust `exp` validation

The AAT constructor range-checked `aud`, but neither token constructor constrained `exp` to the fixed-width range. An out-of-range value encodes at a different CBOR width and silently breaks the fixed offsets the circuits depend on.

## Callers MUST

- expose `trust_anchor_key_x`/`_y` and `now` as **public inputs** — a private `now` lets the prover pick a time at which any token is fresh; a private anchor key lets it sign its own TAKT;
- bind `aat.aud` to the request's RP id, `aat.nonce` to the `ProofRequest` nonce, and `aat.cdh` to the proof's own commitment (zero when nil);
- expose whichever of `sec_flags` and `authenticator_meta` the RP needs.

None of these are enforceable from inside the library. Documented with a worked `main` in `crates/proof/noir/README.md`.


paolodamico:
@codex review

paolodamico:
cursor review

crates/proof/noir/README.md
```diff
@@ -8,6 +8,14 @@
 - `authenticator-attestation/` — WIP-106 Authenticator Attestation library
   (Trust Anchor Key Token + Authenticator Assertion Token verification).
 
+  The TAKT is **not** nested in the AAT. It travels in the AAT's `COSE_Sign1`
```
paolodamico: nit, let's remove the line breaks and delegate to IDEs to render this in the screen

crates/proof/noir/README.md
```diff
@@ -8,6 +8,14 @@
...
+  unprotected header, which never enters the `Sig_structure` (RFC 9052 §4.4), so
+  the circuit never reserializes it. The two tokens are bound solely by the
```
paolodamico:
```suggestion
  the circuit never reserializes it. The two tokens are bound by the
```

crates/proof/noir/README.md
```diff
@@ -8,6 +8,14 @@
...
+  shared `assertion_key`: `verify_takt` authenticates the TAKT under the
+  provider's `trust_anchor_key`, and `verify_aat` then verifies the ES256
+  signature under the key that TAKT attests. That key equality is the *only*
+  binding — if you touch either verification path, preserve it.
```
paolodamico:
```suggestion
```

crates/proof/noir/authenticator-attestation/src/takt.nr
```diff
@@ -29,7 +29,12 @@ pub struct TrustAnchorKeyToken {
 /// freshness is checked (`now < exp`). It MUST be a public input of any outer
 /// proof that calls this function.
 pub fn verify_takt(input: TrustAnchorKeyToken, now: Field) {
-    // CBOR uint32 encoding of `exp` is checked later in `serialize_takt`.
+    // WIP-106 fixes `exp` to a 4-byte CBOR uint. The circuit no longer
```
paolodamico: I'd remove the comment, it's not super informative. furthermore, the last part is a bit misleading because constraint failures aren't surfaced to rust

crates/proof/src/authenticator_attestation/tests.rs
```diff
@@ -290,21 +290,63 @@ fn aat_claims_follow_deterministic_cbor_map_order() {
...
     );
 
-    // submods carries the Trust Anchor Key Token byte-for-byte under "takt".
-    let submods = entries[4].1.as_map().unwrap();
-    assert_eq!(submods.len(), 1);
-    assert_eq!(submods[0].0, Value::Text("takt".to_string()));
-    assert_eq!(submods[0].1, Value::Bytes(takt));
+    // The TAKT is NOT in the payload: it rides in the unprotected header, so it
+    // is outside the AAT signature (WIP-106 AAT section 6).
+    assert!(
```
paolodamico: this seems like a very specific assertion to cover a previous design, I'd drop it, there's multiple things that could be in here that shouldn't. or perhaps a more generic test that ensures only the expected keys are present

paolodamico (dismissed):

crates/proof/noir/authenticator-attestation/src/takt.nr
kilianglas: Yeah, didn't cleanup claude comments yet.

crates/proof/noir/README.md
kilianglas: Done — the README no longer hard-wraps; each bullet and paragraph is a single line. The nested-TAKT paragraph this hunk added is gone entirely too: `authenticator-assertion/` now carries just a one-line WIP warning.

crates/proof/src/authenticator_attestation/tests.rs
```diff
@@ -290,21 +290,63 @@ fn aat_claims_follow_deterministic_cbor_map_order() {
...
+        !entries
+            .iter()
+            .any(|(key, _)| i128::from(key.as_integer().unwrap()) == 266),
+        "submods must not be present in the payload"
+    );
+    let header_takt = sign1
+        .unprotected
+        .rest
+        .iter()
+        .find(|(label, _)| *label == coset::Label::Text("takt".to_string()))
```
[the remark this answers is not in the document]
kilianglas: Moot — that block was removed. The test now pins the claim key set only (`assert_eq!(keys, vec![3, 4, 10, 265, -80_000, -80_001])`), so there is no `coset::Label` comparison left.

crates/proof/src/authenticator_attestation/mod.rs
```diff
@@ -200,6 +200,16 @@ pub enum AttestationError {
     AssertionKeyMismatch,
 }
 
+/// Validates that `exp` fits the fixed 4-byte CBOR uint encoding mandated by
+/// WIP-106 (`2^16 <= exp < 2^32`, dates between 1971 and 2106), so the claim
+/// offsets the circuits rely on stay constant.
```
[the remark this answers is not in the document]
kilianglas: Fixed — the doc comment now reads `2^16 <= exp < 2^32`, i.e. any instant from `1970-01-01T18:12:16Z` up to 2106.

crates/proof/noir/authenticator-attestation/src/aat.nr
```diff
@@ -24,14 +29,15 @@ pub struct AuthenticatorAssertionToken {
     exp: Field,
```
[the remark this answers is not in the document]
kilianglas: Fixed — same correction applied to the `exp` field doc, now at `authenticator-assertion/src/aat.nr` (the package was renamed from `authenticator-attestation`): "any instant from 1970-01-01T18:12:16Z up to 2106".

crates/proof/noir/authenticator-attestation/src/takt.nr
```diff
@@ -29,7 +29,12 @@ pub struct TrustAnchorKeyToken {
...
+    // serializes the TAKT (it rides in the AAT's unprotected header), so this
+    // is the only place that constraint is still enforced -- the Rust
+    // constructor only range-checks the CWT numeric date. Must precede the
+    // freshness check so an out-of-range `exp` reports the encoding error.
```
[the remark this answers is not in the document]
kilianglas: Resolved by deletion — the comment is gone. `verify_takt` now just calls `assert_cbor_uint32(input.exp)` ahead of `assert_not_expired`, so there is no longer any claim about where the invariant lives.

crates/proof/noir/authenticator-attestation/src/takt.nr
kilianglas: Removed — `verify_takt` now just calls `assert_cbor_uint32(input.exp)` before the freshness check, with no commentary. Agreed on the misleading part: constraint failures never surface to Rust.

crates/proof/src/authenticator_attestation/tests.rs
kilianglas: Done — dropped the submods-specific assertions and kept the generic form you suggested: `assert_eq!(keys, vec![3, 4, 10, 265, -80_000, -80_001])` pins exactly the expected claim key set, nothing about a prior design.

dkales (approved):

dkales (approved):

dkales (approved):

PR #885, opened 2026-08-04
feat!: ownership proof (wip-103) correctness improvements

paolodamico:
### Context
Following the Least Authority audit of WIP-103 from July 28, 2026, implementing a number of improvements.

### Circuit changes
_⚠️ These circuit changes imply a breaking change for the entire set of core libraries._

1. Introducing the concept of a `context` which is provided by the verifier. Particularly for the use case of proving `sub` ownership, specifying a specific context to which a proof is bound is useful to reduce the surface area of mis-implemented `nonce` logic. For example, the issuer will request a proof for the `context = b"delete credential"`, preventing accidental use of such proof for e.g. renewing a credential. _We don't expect users to see this context, but it's added as a defense-in-depth mechanism to prevent cross-context use_. This `context` is bound in the authenticator's signature. **Resolution to Suggestion 3**.
2. The `expected_commitment` is now an input instead of an output (as mandated by the spec). This reduces surface area for mis-implementation on the verifier side as the equality between the computed commitment and the expected one is done in the circuit.
3. Removing the dummy constraint for the `nonce`, this isn't actually needed because the `nonce` is already used in the message for the signature.
4. Minor name updates for clarity (e.g. `root` -> `merkle_root`, `commitment_r` -> `commitment_blinder`)

### Other changes
1. Introducing Rust script to re-generate the `Prover.toml` fixture dynamically as well as a full update with real inputs for circuit execution. **Resolution to Suggestion 1**.
2. Updating Rust libraries to reflect the circuit changes.

### Spec Updates
These updates are introduced as a **Resolution to Suggestion 2**.
1. Explicitly call out all inputs, outputs, and all the normative constraints of the circuit, particularly being explicit about the constraints so others can construct a circuit from the spec.
2. Document guidelines and requirements for handling the nonce and the context.

---

> [!NOTE]
> **High Risk**
> Changes cryptographic message hashing, public inputs, and proving/verification APIs across the ownership-proof circuit and core libraries—integrators must regenerate proofs and update verifiers.
> 
> **Overview**
> **Breaking change** to the WIP-103 Ownership Proof circuit and dependent Rust APIs, following Least Authority audit feedback.
> 
> The Noir circuit now takes **`expected_commitment` and `context` as public inputs** (commitment is checked in-circuit instead of being a public output), requires **non-zero `nonce` and `context`**, and signs **`H_4(DS_WIP_103, expected_commitment, nonce, context)`** instead of a 3-element hash. Naming is aligned (`merkle_root`, `leaf_index`, `commitment_blinder`), and the dummy nonce constraint is removed. The circuit gains extensive negative tests; CI adds **`nargo execute --pedantic-solving`** for ownership-proof.
> 
> Rust **`prove_credential_sub`** and **`verify_ownership_proof`** take **`context`**; proving uses shared **`message_digest`**, pre-prove **`check_ownership_input_validity`**, and drops direct **`poseidon2`** from the authenticator crate. **`Prover.toml`** is generated from a shared fixture via **`UPDATE_PROVER_TOML=1`** (workspace **`toml`** dep).
> 
> **WIP-103** is expanded with explicit inputs, constraints, and nonce/context rules for verifiers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bdfd57df46b970d0fa87f7b9519d0aabd1afc692. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


paolodamico:
@codex review

docs/WIPs/wip-103.md
```diff
@@ -33,61 +33,121 @@ The terms [Authenticator](https://docs.rs/world-id-core/latest/world_id_core/str
...
-4. Unless otherwise noted all hashing required by this spec must be using the `Poseidon2` function and signature using Edwards-curve Digital Signature Algorithm (`EdDSA`) over the `BabyJubJub` curve, in compliance with the World ID Protocol.
-5. Recipients of an **Ownership Proof** (also called "Verifiers" in this spec) MUST enforce one-time use of nonces. Verifiers are responsible for providing the nonce.
-6. If future use cases are added, the `blinder` IS REQUIRED to always be a high entropy value to prevent finding a user's `leafIndex`. Unless otherwise noted, it should be randomly generated from a CSPRNG.
+A **non-normative high-level description** of what the circuit proves:
+1. It proves the user's `leaf_index` hashed with a `commitment_blinder` and a specific domain separator produces an `expected_commitment`. _An example, is the Credential's blinded `sub`_.
+2. It proves the user's `leaf_index` is validly registered in the `WorldIDRegistry` and the user knows the secret key ($sk$) to a registered authenticator for the particular `leaf_index`.
```
dkales: This sk knowledge is technically not proven, just via the proxy of the signature.

dkales (approved):

docs/WIPs/wip-103.md
```diff
@@ -33,61 +33,121 @@ The terms [Authenticator](https://docs.rs/world-id-core/latest/world_id_core/str
...
+
+Every occurrence of $H$ in the constraints below refers to $H_t$ for the stated $t$.
+
+### Circuit constraints
+
+The Ownership Proof circuit MUST implement the following constraints (order is not relevant):
+
+1. **Expected commitment**. Compute the commitment as: `H_3(domain_separator, leaf_index, commitment_blinder)`. The computed commitment MUST equal the `expected_commitment`. The `domain_separator` is dependent on the [Use Case](#use-cases) and all possible domain separators MUST be hardcoded in the circuit.
+2. **Signature verification**. Compute the message of the signature as: `message = H_4(DS_WIP_103, expected_commitment, nonce, context)`. The circuit MUST verify `(query_r, query_s)` as an EdDSA-Poseidon2 signature by `pk` over the `message` in order to prove both knowledge of `sk` and authorization over the particular `nonce` and `context`. With $`R = (\mathtt{query\_r[0]}, \mathtt{query\_r[1]})`$ and $`s = \texttt{query\_s}`$, the circuit MUST constrain:
+    1. **Canonical scalar**: $s < L$ to prevent signature malleability (i.e. reduced scalar).
```
dkales: L is never defined (it is the order of the large prime-order subgroup the BabyJubJub Elliptic Curve)

docs/WIPs/wip-103.md
```diff
@@ -33,61 +33,121 @@ The terms [Authenticator](https://docs.rs/world-id-core/latest/world_id_core/str
...
+    2. **$R$ on curve**. $R$ must be a point satisfying the BabyJubJub curve equation because addition is only complete for on-curve points. The point is NOT REQUIRED to lie in the prime-order subgroup.
+    3. **Challenge**. $`h = H_8(\texttt{DS\_EDDSA}, R_x, R_y, A_x, A_y, \mathtt{message})`$, so `h` is used as a full-field scalar without reduction.
+    4. **Cofactored verification equation**. The signature MUST meet the cofactored verification equation.
```
dkales: Maybe this is nice as a reference: https://eprint.iacr.org/2020/1244.pdf

dkales (dismissed):

docs/WIPs/wip-103.md
```diff
@@ -33,61 +33,121 @@ The terms [Authenticator](https://docs.rs/world-id-core/latest/world_id_core/str
...
+5. **Leaf index commitment**. The circuit MUST compute `offchainSignerCommitment` as designated in the Query Proof, with $H_{16}$ using the domain separator `MERKLE_LEAF_DS` and each key in `user_pk` encoded as $(x, y)$ points sequentially, with the last position (index `15`) equal to `0`. The digest is output state element `1`. All keys MUST be included in order, including _empty_ slots with the neutral element $O$.
+6. **Merkle inclusion proof**. The circuit MUST recompute `merkle_root` from `offchainSignerCommitment` as the leaf value at position `leaf_index`, with:
+    1. `depth <= MAX_DEPTH`. A `depth` greater than `MAX_DEPTH` MUST cause proving to fail.
+    2. The $`\texttt{node}_0 = \texttt{offchainSignerCommitment}`$ and for $`i \in [0, \texttt{depth})`$, $`\texttt{node}_{i+1} = C(left_i, right_i)`$
```
dkales: While it says "at positoin leaf index" above, this could use a bit more exposition that the node_i is either the left or right input based on the i-th bit of leaf_index, and the other node is the one from the merkle path.

crates/authenticator/src/prove.rs
```diff
@@ -477,6 +477,8 @@ impl Authenticator {
     ///
     /// # Arguments
     /// - `nonce`: The nonce of the request provided by the Issuer.
+    /// - `context`: The context provided by the Issuer for which the proof is valid (e.g. "delete a credential"). The Issuer
```
Dzejkop: maybe `context_hash` or `context_commitment`? the comment here lightly implies we might want this to be a string of sorts

crates/proof/noir/ownership-proof/src/commitment.nr
```diff
@@ -28,3 +28,32 @@ fn test_generate_commitment_different_inputs() {
...
     assert(msg1 != msg3);
 }
+
+#[test]
+fn test_generate_commitment_is_not_symmetric() {
+    assert(generate_commitment(1, 100) != generate_commitment(100, 1));
+}
+
+#[test]
+fn test_generate_message_deterministic() {
```
Dzejkop: I get the idea - but I don't think this test actually proves anything

crates/proof/noir/ownership-proof/src/types.nr
```diff
@@ -1,24 +1,32 @@
...
-    pub mt_index: Field,
+    /// The index of the user's World ID in the `WorldIDRegistry`, it is the
+    /// position it holds in the Merkle tree. In other circuits this is called `mt_index`.
+    pub leaf_index: Field,
+    /// The sequence of all siblings from the `leaf_index` to the `merkle_root`.
     pub siblings: [Field; MAX_DEPTH],
 }
 
+/// The private inputs of the Ownership Proof (WIP-103). The public inputs are declared
+/// directly on `main`.
```
Dzejkop:
```suggestion
/// The private inputs of the Ownership Proof (WIP-103).
```

crates/proof/src/fixtures.rs
```diff
@@ -0,0 +1,206 @@
...
+        inclusion_proof: MerkleInclusionProof::new(root, LEAF_INDEX, siblings),
+        nonce,
+        expected_commitment,
+        context,
+        signature: sk.sign(*message_digest(expected_commitment, nonce, context)),
+        commitment_blinder,
+    }
+}
+
+mod ownership_proof_prover {
```
Dzejkop, line 43: you could probably use [quote](https://crates.io/crates/quote) here and then `.to_string()` the output

Dzejkop: ok, quote obviously will not work here, my bad

but consider https://github.com/worldcoin/world-id-protocol/pull/906

crates/proof/noir/ownership-proof/src/commitment.nr
kilianglas: Wouldn't it be better to replace this by a known output test?

crates/proof/noir/ownership-proof/src/types.nr
```diff
@@ -1,24 +1,32 @@
...
 pub struct OwnershipProofInputs<let MAX_DEPTH: u32, let NUM_KEYS: u32> {
```
kilianglas:
```suggestion
pub struct OwnershipProofPrivateInputs<let MAX_DEPTH: u32, let NUM_KEYS: u32> {
```

kilianglas (dismissed):

crates/authenticator/src/prove.rs
```diff
@@ -466,6 +466,10 @@ impl Authenticator {
     ///
     /// # Arguments
     /// - `nonce`: The nonce of the request provided by the Issuer.
+    /// - `context`: The field element the Issuer assigns to the operation being authorized (e.g. one value for
+    ///   credential deletion, a different one for renewal). The Issuer defines the encoding, which may be arbitrary
```
0xOsiris: > The Issuer defines the encoding, which may be arbitrary bytes reduced into the field

As stated this would make collisions constructible > 31 bytes right? Context should probably rather have a prescribed encoding of something like e.g. `keccak256("operation")`

0xOsiris (dismissed):

crates/authenticator/src/prove.rs
paolodamico: technically possible to have collisions if > modulus and you don't use a hash function to smooth things out but naturally a very remote case. thoughts on having a RECOMMENDED guideline on this vs. enforcing eg. length?

0xOsiris (approved):

PR #887, opened 2026-08-05
refactor(noir): single entrypoint for attestation verification

kilianglas:
Makes the attestation library consumable, and makes it impossible to consume unsafely. Stacked on #884 — review that one first.

## The problem

Two things were wrong with the API, both surfaced while sketching how the DeepFace circuits would call it.

**The library could not be used at all.** Every field of `TrustAnchorKeyToken` and `AuthenticatorAssertionToken` was module-private, so no external Noir package could construct either token. I confirmed this with a throwaway consumer package: `error: exp is private and not visible from the current module`, and the same for all eight fields. `verify_aat` was reachable but uncallable.

**The binding was carried by struct nesting.** Since #884 un-nested the TAKT, the shared `assertion_key` is the *sole* thing tying the two tokens together — the spec says so explicitly ("This equality is the sole binding between the two tokens"). That made it more important, not less, that neither token can be verified in isolation. But the AAT held the TAKT as a field, which mirrors the old nested wire format rather than the current one.

## The shape

```rust
verify_takt(takt, trust_anchor_key_x, trust_anchor_key_y, now)   // pub(crate)
verify_aat(aat, assertion_key_x, assertion_key_y, now)           // pub(crate)
verify_attestation(aat, takt, trust_anchor_key_x, _y, now)       // pub
```

The composite feeds `verify_aat` the `assertion_key` read from the TAKT it just verified. The binding therefore holds *by construction* — no equality assert, no second witness for the same key, nothing a refactor can drop. And because the primitives are crate-private, a consumer cannot verify one half and skip the other. From the same external package:

```
error: verify_aat is private and not visible from the current module
error: verify_takt is private and not visible from the current module
```

That is the bypass — verify an AAT under a prover-chosen key, never check a TAKT — failing at compile time.

Each primitive is also independently testable now. Previously every AAT test had to drag in a full valid TAKT fixture.

## Two structural fixes that fall out

**The AAT no longer carries the TAKT, and the TAKT no longer carries the `trust_anchor_key`.** Neither was part of the token it sat in. The TAKT travels in the AAT's unprotected header. And the trust anchor key is not a TAKT claim — the token identifies its signer only by the `kid` protected-header parameter, which WIP-106 leaves outside the signature and the Security section says to treat as an untrusted hint. It is a verifier-supplied input, exactly like `now`; `takt_digest` never hashed it. The signature now reflects that.

**Struct fields are `pub`**, so tokens can be constructed. This is the fix for "not consumable," and it is orthogonal to the guarantee above: the prover supplies the token witnesses either way — what matters is that it cannot skip a verification step.

## Tests

- `attestation::test_untrusted_trust_anchor_key_rejected` — new. A well-formed TAKT signed by anyone other than the RP's allowlisted key must fail. Checked against the BabyJubJub prime-order subgroup generator so the failure is the signature check, not a point-validity assert (an off-curve key trips `Public key must be on curve` first, which would have made the test pass for the wrong reason).
- `test_aat_key_not_attested` moved to the composite, where the binding now lives.
- `aat::test_aat_wrong_assertion_key` — the primitive-level counterpart.

## What the library still cannot enforce

Documented in `crates/proof/noir/README.md` with a worked `main` example, because no API shape can enforce it: the calling circuit must expose `trust_anchor_key` and `now` as **public inputs** (a private `now` lets the prover pick a time at which any token is fresh; a private anchor key lets it sign its own TAKT), must bind `aat.nonce`/`aat.cdh` to the surrounding proof, and must expose whichever of `sec_flags` and `authenticator_meta` its RP needs. A CI check over each consumer's compiled ABI would catch the public-input half; worth adding alongside the first consumer.

## Test plan

- [ ] `cd crates/proof/noir/authenticator-attestation && nargo test` — 14 passing (was 11)
- [ ] `cargo test -p world-id-proof authenticator_attestation` — 16 passing, unchanged (the Rust encoder is untouched)
- [ ] `nargo fmt --check` clean
- [ ] External consumer package compiles against `verify_attestation`, with `trust_anchor_key_x`/`_y` and `now` reported as `public` in the emitted ABI
- [ ] Same package fails to compile when calling `verify_aat` or `verify_takt` directly

### DIFF
--- a/crates/proof/noir/README.md
+++ b/crates/proof/noir/README.md
@@ -16,6 +16,39 @@
   signature under the key that TAKT attests. That key equality is the *only*
   binding — if you touch either verification path, preserve it.
 
+## Consuming `authenticator-attestation`
+
+`attestation::verify_attestation` is the only entrypoint. `verify_takt` and
+`verify_aat` are crate-private by design: a TAKT alone attests a key without
+saying what it signed, and an AAT alone proves only that *some* key signed the
+claims. Going through the composite is what binds them — it feeds `verify_aat`
+the `assertion_key` read from the TAKT it just verified, so the two cannot be
+mismatched. If you find yourself wanting one half, verify both and ignore what
+you don't need.
+
+The library cannot enforce the rest. A calling circuit MUST:
+
+- expose `trust_anchor_key_x`/`_y` and `now` as **public inputs**. The trust
+  anchor key is the RP's entire trust decision, taken from the provider's
+  Authenticator Metadata; left private, a prover signs its own TAKT and the
+  attestation proves nothing. A private `now` lets a prover pick a time at which
+  any token is fresh.
+- bind `aat.nonce` and `aat.cdh` to the surrounding proof.
+- expose whichever of `sec_flags` (via `takt::unpack_sec_flags`) and
+  `aat.authenticator_meta` the RP needs for its business rules.
+
+```rust
+fn main(
+    trust_anchor_key_x: pub Field,
+    trust_anchor_key_y: pub Field,
+    now: pub Field,
+    aat: AuthenticatorAssertionToken,
+    takt: TrustAnchorKeyToken,
+) {
+    verify_attestation(aat, takt, trust_anchor_key_x, trust_anchor_key_y, now);
+}
+```
+
 ## Development workflow
 
 Circuits are developed and tested with **nargo only** — no proving backend:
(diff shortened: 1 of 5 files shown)


kilianglas:
Folding this into #884 instead of stacking another PR — commit `0d65c97` now sits on `kilianglas/aat-unnest-takt`, rebased onto the two README suggestion commits. Closing.

PR #888, opened 2026-08-06
feat: WIP-100 Cryptographic Primitives for the Protocol

paolodamico:
Base cryptographic primitives for the World ID Protocol. WIP-100 will be a living spec, updated over time.


dkales (approved):

0xForerunner (approved):
Looks good! This will be really nice to have when making future changes to the protocol. As far as I can tell this lines up pretty well with the current protocol implementation.

PR #892, opened 2026-08-07
fix: set claims hash capacity

paolodamico:
### Motivation
The `claims_hash` didn't have a capacity element which would allow for finding second pre-images. This is not a practical issue today because the credential's `claims` are reserved for future use, and in practice no known issuer is pre-registering more than 5 claims, but this is unsound and could affect the claims' guarantees if it were in use. This change is added to prevent future issues using claims.

### Changes
Explicitly sets the 15th-index of the `claims_hash` as the sponge's capacity (using `0` as value). 

### Future Improvements
In the next credential version, the capacity should be added with an explicit ds at element[0] following Protocol conventions.

---

> [!NOTE]
> **Medium Risk**
> Changes the cryptographic definition of `claims_hash` for V1 credentials; impact is limited today because claims are unused in production proofs, but any issuer or test relying on the old 16-slot hash or index 15 would diverge.
> 
> **Overview**
> Fixes **unsound V1 `claims_hash` computation** by reserving Poseidon2 `t=16` **element[15] as sponge capacity (0)** instead of treating all 16 slots as claim inputs, which could allow second preimages once claims are used in proofs.
> 
> **`Credential::MAX_CLAIMS` drops from 16 to 15**; `claim` / `claim_hash` reject index 15 and oversized claim vectors. **`claims_hash`** builds a 16-word state (claims in `0..14`, zero capacity at 15), returns **`PrimitiveError`** on overflow, and serde **`deserialize_claims`** enforces the limit while **stripping a legacy trailing zero 16th claim** so old serialized credentials still verify.
> 
> **`world_id_proof`** surfaces credential errors via **`ProofError::CredentialError`** when assembling nullifier circuit inputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7382c9bba3b3bf9725b0829f7232052e8fd7435e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/primitives/src/credential.rs
+++ b/crates/primitives/src/credential.rs
@@ -22,7 +22,7 @@ pub enum CredentialVersion {
     /// - Curve (Base) Field (`Fq`): `BabyJubJub` Curve Field (also the BN254 Scalar Field)
     /// - Scalar Field (`Fr`): `BabyJubJub` Scalar Field
     ///
-    /// **NOTE**: In V1, the `claims_hash` uses the last element[15] of the Poseidon2
+    /// **NOTE**: In V1, the `claims_hash` uses the last element (`element[15]`) of the Poseidon2
     /// permutation as the capacity (value = `0`).
     #[default]
     V1 = 1,
@@ -153,6 +153,9 @@ pub struct Credential {
     /// the value directly (e.g. encoded date of birth).
     ///
     /// Currently these statements are not in use in proofs yet.
+    ///
+    /// Deserialization accepts at most [`Credential::MAX_CLAIMS`] claims.
+    #[serde(deserialize_with = "deserialize_claims")]
     pub claims: Vec<FieldElement>,
     /// The commitment to the Associated Data issued with the Credential.
     ///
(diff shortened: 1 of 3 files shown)


paolodamico:
@codex review

paolodamico:
cursor review

dkales (dismissed):

dkales (dismissed):

crates/primitives/src/credential.rs
```diff
@@ -324,16 +334,18 @@ impl Credential {
...
-            eyre::bail!("There can be at most {} claims", Self::MAX_CLAIMS);
+            return Err(PrimitiveError::OutOfBounds);
         }
-        let mut input = [*FieldElement::ZERO; Self::MAX_CLAIMS];
+        let mut input = [*FieldElement::ZERO; Self::MAX_CLAIMS + 1]; // +1 is the capacity value
         for (i, claim) in self.claims.iter().enumerate() {
             input[i] = **claim;
         }
+
+        debug_assert_eq!(input[15], *FieldElement::ZERO);
```
0xOsiris, line 347: This seems a bit tautological with the above

```rust
        let mut input = [*FieldElement::ZERO; Self::MAX_CLAIMS + 1]; // +1 is the capacity value
```

crates/proof/src/lib.rs
```diff
@@ -65,6 +65,9 @@ pub enum ProofError {
     /// Error verifying a Noir Proof with ProveKit. This usually means the proof is invalid.
     #[error("proof verification error: {0}")]
     Verification(String),
+    /// The proof cannot be generated because the credential has an error
+    #[error("credential error: {0}")]
+    CredentialError(String),
```
0xOsiris:
```suggestion
    #[error(transparent)]
    CredentialError(#[from] PrimitiveError),
```

0xOsiris (dismissed):

dkales (approved):

crates/primitives/src/credential.rs
paolodamico: it is absolutely tautological, it is added for an abundance of clarity and to avoid accidental overwrites

Dzejkop (approved):

kilianglas (approved):

PR #908, opened 2026-08-11
refactor(authenticator): move session r-seed verification to SessionId

kilianglas:
* Moves `Authenticator::validate_cached_session_r_seed` to `SessionId::verify_commitment`, which re-derives a `SessionId` from an `r` seed and compares commitments
* No behavior change. Same comparison, same `AuthenticatorError::SessionIdMismatch` on mismatch.

Addresses [#855 (comment)](https://github.com/worldcoin/world-id-protocol/pull/855#discussion_r3597433054) and [#855 (comment)](https://github.com/worldcoin/world-id-protocol/pull/855#discussion_r3595502137).


crates/primitives/src/session.rs
```diff
@@ -156,6 +156,22 @@ impl SessionId {
...
+    /// Checks an `r` seed obtained elsewhere (e.g. an Authenticator's cache) without
+    /// re-deriving it via the OPRF nodes.
+    ///
+    /// # Errors
+    /// If this session id's [`Self::oprf_seed`] is not valid.
+    pub fn verify_commitment(
+        &self,
+        leaf_index: u64,
+        session_id_r_seed: FieldElement,
+    ) -> Result<bool, PrimitiveError> {
```
dkales: nit: I find `Result<bool>` to be a bit error prone, since there is no way to enforce must_use on the bool.
so something like `comm.verify_commitment(1,2)?;` looks like it does verify it, but actually does not

paolodamico: +1, easy to miss

crates/authenticator/src/prove.rs
```diff
@@ -204,7 +204,7 @@ impl Authenticator {
     pub async fn build_session_id(
```
paolodamico, line 204: would be great to have unit tests on this too. there's no test coverage to ensure `request_session_id.verify_commitment` is actually run

Dzejkop (dismissed):

dkales (approved):

paolodamico (approved):

PR #934, opened 2026-08-14
feat: RP Request EIP712 Refactor

0xForerunner:
V1's RP signature covers only `version || nonce || created_at || expires_at || [action]`, so it authenticates a time window rather than authorizing an operation. This leaves `rp_id`, `proof_type`, `session_id`, the credential items, and the constraint tree entirely unbound. 

The main issue is that a plain uniqueness signature can be replayed as a session-seed query via the `rp_signature_verification`, letting a client create a session the RP never authorized. V2 replaces this with an EIP-712 payload that binds an explicit operation matrix plus a salted detailsHash commitment to the private request body. The salting is necessary because OPRF nodes must verify the authorization but should not learn what was requested.

---

> [!NOTE]
> **High Risk**
> Breaking change to RP authorization and OPRF node verification across the proof path; mis-deployed clients, nodes, or WIP-101 contracts could reject valid requests or weaken binding between signed intent and OPRF operations.
> 
> **Overview**
> **Replaces the legacy RP message signature** with **EIP-712 typed authorization** that binds the full public operation matrix (`proof_type`, `session_mode`, action, `rp_id`, `oprf_key_id`, timestamps, nonce) plus a **`detailsHash`** commitment to private request fields (items, constraints, session ref, `request_id`). **`ProofRequest` now requires a non-zero `request_salt`** (RP–Authenticator only; not sent to OPRF nodes).
> 
> **OPRF nullifier auth** is reshaped: duplicated RP metadata is collapsed into **`RpRequestAuthorization`**, the evaluated input is **`oprf_action`**, and proof material is an explicit **`RpAuthorizationProof`** (`Eoa` signature vs **`Wip101`** calldata). Session-seed rederivation can include **`session_seed_opening`**. Error **`InvalidRpSignatureVerification`** becomes **`InvalidRpAuthorization`** (code `4524` unchanged).
> 
> **Authenticator / proof pipeline** gains **`generate_*_with_authorization`** so contract-backed RPs can pass WIP-101 proof through nullifier, session-seed, and full proof flows. **WIP-101** and mocks now use **`verifyRpRequest(RpRequest intent, uint256 oprfAction, bytes data)`** aligned with the same `RpRequest` struct EOAs sign. **OPRF node** resolves **chain id at startup** (async **`start`**) for the EIP-712 domain; dev client and e2e tests sign via **`eip712_signing_hash`**.
> 
> This is a **coordinated breaking rollout** (clients, nodes, WIP-101 contracts); legacy signatures and the old WIP-101 calldata are dropped. Spec/docs added under **`rp-request-authorization.md`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f8dd6587ee0971c56cff2c6120770da6c53209f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/oprf-node/src/auth/rp_module/wip101/tests.rs
```diff
@@ -1,243 +1,272 @@
...
+    bytes4 constant ERC165_INTERFACE_ID = type(IERC165).interfaceId;
+    bytes4 constant IWIP101_INTERFACE_ID = type(IWIP101).interfaceId;
 
-#[tokio::test]
-async fn verify_custom_error() {
-    let auth = dummy_auth();
-    let revert = IWIP101::RpInvalidRequest {
-        code: U256::from(42),
+    #[sol(rpc, bytecode="60808060405234601557610192908161001a8239f35b5f80fdfe6080806040526004361015610012575f80fd5b5f3560e01c90816301ffc9a7146100d6575063d7e3f32c14610032575f80fd5b346100d2577ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffc36016101c081126100d257610180136100d2576101a43567ffffffffffffffff81116100d257366023820112156100d257806004013567ffffffffffffffff81116100d257369101602401116100d25760206040517fd7e3f32c000000000000000000000000000000000000000000000000000000008152f35b5f80fd5b346100d25760207ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffc3601126100d257600435907fffffffff0000000000000000000000000000000000000000000000000000000082168092036100d257817fd7e3f32c0000000000000000000000000000000000000000000000000000000060209314908115610168575b5015158152f35b7f01ffc9a7000000000000000000000000000000000000000000000000000000009150148361016156")]
```
0xThemis: is there a reason we revert this change?

imo the mock asserter tests are nicer then we dont need to maintain these mock contracts and need to spawn anvil machinery on every test

services/oprf-node/src/lib.rs
```diff
@@ -102,6 +104,16 @@ pub fn start(
             .build()
             .context("while init blockchain connection")?;
 
+    // The chain is a binding parameter of the RP request EIP-712 domain, so resolve it once
+    // here instead of on the request path.
+    let chain_id = tokio::time::timeout(
```
0xThemis, line 109: nit: dont think we need to wrap that into timeout. We only added that because we call into arbitrary defined contracts registered by untrusted RPs but getting the chain_id from the provider should always be fine

services/oprf-node/src/auth/rp_module/wip101/tests.rs
0xForerunner: No apologies this was a rebasing issue. I will revert this.

kilianglas (commented):
I'd prefer to move `RpRequestAuthorization` and related logic into a `request_authorization.rs` module/ file under `request/` instead of having it in `rp.rs`

crates/primitives/src/oprf.rs
kilianglas: NIT: We could rename `action` to `oprf_action` to indicate that this is not necessarily the same action as in the RP authorization (e.g., for a session seed request that is sent as part of a bound uniqueness proof request)

kilianglas: `nonce`, `created_at`, `epires_at` and `rp_id` are duplicate now. We could remove those fields and only read them from `rp_request_authorization` instead. Suggested definition:
```rust
    pub struct NullifierOprfRequestAuthV1 {
    pub proof: Proof<Bn254>,
    pub oprf_action: ark_babyjubjub::Fq,
    pub merkle_root: ark_babyjubjub::Fq,
    pub authorization: RpRequestAuthorization,
    pub authorization_proof: RpAuthorizationProof,
    pub session_seed_opening: Option<B256>,
}
```

kilianglas:
I don't fully understand why we need the private part of the request authorization. OPRF nodes are the only place where rp request signatures are currently verified, so imo the RP should only sign over fields the OPRF nodes need to check. Or am I missing something here?

crates/primitives/src/rp.rs
```diff
@@ -2,15 +2,384 @@
...
+    ///
+    /// # Errors
+    /// Returns an error if the request is not structurally valid.
+    pub fn eip712_signing_hash(
+        &self,
+        chain_id: u64,
+        rp_registry: Address,
+    ) -> Result<B256, PrimitiveError> {
+        Ok(self.rp_authorization()?.signing_hash(chain_id, rp_registry))
+    }
```
kilianglas: NIT: I'd prefer having a TryInto instance in the same file the `RpRequestAuthorization` is in and then an `rp_authoriazation` method that just calls `try_into` in `request/mod.rs`

docs/world-id-4-specs/rp-request-authorization.md
```diff
@@ -0,0 +1,104 @@
+# RP request authorization
```
kilianglas, line 1: This entire file seems to be a bit verbose. Maybe we could even merge this into the main spec?

PR #920, opened 2026-08-13
refactor(gateway): extract Redis request store

Dzejkop:
Extracts the Redis `request` store as a facade to its own module complete with extra doc comments + unit tests

---

> [!NOTE]
> **Medium Risk**
> Changes duplicate-in-flight locking and terminal lock release semantics in shared Redis state used across gateway replicas; well-tested but still concurrency-sensitive.
> 
> **Overview**
> Moves gateway Redis persistence (request records, pending set, in-flight locks, rate limits, Lua scripts, and TTLs) out of `request_tracker` into a dedicated **`request_store`** module. **`RequestTracker`** now delegates to **`RequestStore`** and keeps HTTP/metrics/backlog behavior.
> 
> **Behavioral tweaks** bundled with the move: in-flight lock values are the **owning request ID** (not literal `1`), create rollback only clears locks it just set, and terminal status cleanup deletes a lock only when the owner matches that request or the legacy `1` value—covering re-acquired locks after TTL expiry.
> 
> Adds **`indoc`** for a serialization snapshot test and **Redis testcontainer** integration tests for lifecycle, duplicate locks, legacy locks, and rate limiting. **`Cargo.lock`** picks up transitive `windows-sys` version bumps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb4f798b2d0f0d052e146086eb67c474ed3bb83d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #935, opened 2026-08-17
refactor(services): normalize provider configuration

Dzejkop:
## Summary

- replace nested optional provider, signer, and retry configuration with default values
- update service and test callers to use the normalized configuration
- move the gateway request store under the storage module

## Stack

Previous: #920

Next: #936

## Validation

- `cargo +nightly fmt --all -- --check`
- focused common, gateway, and indexer tests
- focused Clippy with warnings denied

---

> [!NOTE]
> **Medium Risk**
> Touches shared RPC provider construction and gateway startup validation used by transaction submission; behavior should be equivalent but misconfigured deploys could fail differently (empty RPC list vs missing signer).
> 
> **Overview**
> **Provider configuration** is flattened so `http`, `signer`, and `retry` are always present (empty vec / default structs) instead of `Option` wrappers. RPC URL presence is checked with `http.is_empty()`; optional signing is detected via `signer.signer_config()` rather than `signer: None`. Builder helpers and `http_with_nonce_manager` were updated accordingly, and `SignerArgs` now derives `Default`.
> 
> **Gateway** validates that a signer is configured using `signer_config()`, and the Redis **request store** lives under a new `storage` module (`request_tracker` imports were repointed).
> 
> **Tests and tools** that construct `ProviderArgs` / gateway configs were updated to pass `http` and `signer` directly (e.g. `SignerArgs::default()` where no wallet is needed).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68bf17dc2b79c1468656a8e5cf3a9f624bf19c2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/common/src/alloy/provider.rs
```diff
@@ -309,14 +307,15 @@ impl ProviderArgs {
...
-        let Some(http) = self.http else {
+        if self.http.is_empty() {
             return Err(ProviderError::NoHttpUrls);
-        };
+        }
 
         // Save first URL for signer (needed for AWS KMS chain_id lookup)
-        let first_url = http.first().cloned().ok_or(ProviderError::NoHttpUrls)?;
+        let first_url = self.http[0].clone();
+        let http = self.http;
```
piohei, line 316: Do we need this actually? Can we just use `self.http` in the code?

piohei (dismissed):

services/common/src/alloy/provider.rs
Dzejkop: it's kind of a workaround I think - I'd leave it for now until https://github.com/alloy-rs/alloy/issues/4150 is resolved

piohei (approved):

PR #941, opened 2026-08-17
docs: state the trust anchor obligation on the WIP-106 verifiers

kilianglas:
Adds comments for consistency reasons. Addresses Finding 5 in https://hackmd.io/@m1guelpf/protocol-daybreak.


kilianglas:
Superseded by #884, which fixes this structurally rather than by documentation: `trust_anchor_key_x`/`_y` are removed from the `TrustAnchorKeyToken` witness struct and become explicit parameters of the crate-root `verify_attestation`, `verify_takt`/`verify_aat` become `pub(crate)` so the obligation cannot be bypassed via the inner helpers, and the public-input requirement is documented on the entrypoint plus `crates/proof/noir/README.md` with a worked `main` example. `test_untrusted_trust_anchor_key_rejected` pins it.

This PR only documented the obligation, and its edits sit in doc comments that #884 rewrites wholesale, so merging it would just create a conflict. Closing unmerged.

PR #141, opened 2025-12-04
feat: proof requests & responses (take II)

paolodamico:
> [!NOTE]
> Introduce `requests` API (proof requests/responses with constraints) and update authenticator to generate proofs from `ProofRequest`, with supporting primitives, tests, and stubs changes.
> 
> - **Core**:
>   - **Requests API**: Add `requests` module with `ProofRequest`, `RequestItem`, `ProofResponse`, constraint AST (`all`/`any`), versioning, JSON (de)serialization, `digest_hash`, validation, and selection helpers.
>   - **Authenticator**: Change `generate_proof` to accept `ProofRequest` + `Credential`; select request by `issuer_schema_id` and use per-item `signal_hash`; add `InvalidCredentialForProofRequest` error.
>   - **CLI**: Update `bin/authenticator.rs` to read `ProofRequest` from file.
>   - **Exports**: Expose `requests` in `lib.rs`.
> - **Primitives**:
>   - Add `FieldElement::from_arbitrary_raw_bytes`; derive `PartialEq` for `WorldIdProof`; minor `Credential` doc tweak.
> - **Tests/Utils**:
>   - Rewrite e2e/unit tests for new requests/constraints; update fixtures/stubs (nonce/timestamp encoding, RP fixture), and add `ark-serialize` dep.
> - **Cleanup**:
>   - Remove legacy `types::RpRequest`; adjust Cargo features (enable `k256` serde).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f3e5911dcdf10ae8bdac45edbe46ea992079a7b7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #955, opened 2026-08-27
feat: tolerate unknown JSON fields in proof request/response envelopes

SeanROlszewski:
### Problem

`ProofRequest`, `RequestItem`, `ProofResponse`, and `ResponseItem` all carry `#[serde(deny_unknown_fields)]`, which makes every minor protocol addition a breaking change for deployed counterparties. An RP that adds a field to a request breaks every older Authenticator, and an Authenticator that adds a field to a response breaks every older RP.

I want Authenticators to optionally disclose credential `claims` on response items so relying parties can consume them, bound into the mobile app-integrity signature. With strict parsing, emitting any new response field breaks all existing verifiers with a `serde` error, so the field can never be rolled out incrementally.

As far as I can tell the strictness was not a documented protocol decision — the attribute has been on these structs unchanged and uncommented since they were introduced in #141, with no mention in the PR, review threads, or the specs. If it *was* deliberate, I'd love to hear the reasoning; this PR is as much a design question as a change.

### What deliberately stays strict

`ConstraintExpr` / `ConstraintNode` keep `deny_unknown_fields`: they are `#[serde(untagged)]` enums where the attribute is load-bearing for variant disambiguation — without it, `{"all": [...], "any": [...]}` would silently match `All` and drop the `any` branch, changing constraint semantics. That is a correctness guard, not forward-compat friction.

### Solution

* Remove `deny_unknown_fields` from the four envelope structs and document the forward-compatibility policy on each.
* Add tests pinning the tolerance at every nesting level for both requests and responses.

### Testing

* `cargo test -p world-id-primitives --lib` — 167 passed, including the two new tolerance tests.
* `cargo clippy -p world-id-primitives` — clean.

---

> [!NOTE]
> **Medium Risk**
> This is a wire-format compatibility change for core proof envelopes; behavior is safer for signatures but loosens parsing, so downstream code must not assume unknown fields imply rejection.
> 
> **Overview**
> **Relaxes JSON parsing** on `ProofRequest`, `RequestItem`, `ProofResponse`, and `ResponseItem` by dropping `#[serde(deny_unknown_fields)]`, so minor protocol additions no longer break older Authenticators (extra request fields) or older RPs (extra response fields such as future `claims`).
> 
> Doc comments on each struct state that unknown keys are ignored on deserialize and that RP signing / `digest_hash` still cover only enumerated fields, not extra JSON.
> 
> **Tests** `request_json_tolerates_unknown_fields` and `response_json_tolerates_unknown_fields` assert parsing succeeds with unknown keys at the root and inside nested `proof_requests` / `responses` items. `ConstraintExpr` / `ConstraintNode` are unchanged and remain strict.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 302f2c1c3e8673f5491eef3c15ffa53bfae81816. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/primitives/src/request/mod.rs
+++ b/crates/primitives/src/request/mod.rs
@@ -85,8 +85,13 @@ impl ProofType {
 }
 
 /// A proof request from a Relying Party (RP) for an Authenticator.
+///
+/// Unknown JSON fields are ignored during deserialization so that older
+/// Authenticators keep accepting requests from RPs speaking a newer minor
+/// revision of the protocol. The RP signature only covers the enumerated
+/// fields (see [`ProofRequest::digest_hash`]), so tolerated fields are
+/// never part of any signed or hashed message.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
 pub struct ProofRequest {
     /// Unique identifier for this request.
     pub id: String,
@@ -135,8 +140,10 @@ pub struct ProofRequest {
 }
 
 /// Per-credential request payload.
+///
+/// Unknown JSON fields are ignored during deserialization; see
+/// [`ProofRequest`] for the forward-compatibility rationale.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
 pub struct RequestItem {
     /// An RP-defined identifier for this request item used to match against constraints and responses.
     ///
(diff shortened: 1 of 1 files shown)


Takaros999 (approved):
lgtm

Dzejkop (approved):

0xOsiris (approved):

PR #958, opened 2026-08-27
feat: disclose credential claims on RP opt-in

SeanROlszewski:
### Problem

Relying parties need access to the claims of the credential used in a World ID 4.0 proof (e.g. face-credential claims consumed by the wallet bridge verifier and bound into the mobile app-integrity signature — worldcoin/oxide#1178). Today the proof response carries no claims, and there is no way for an RP to ask for them.

Stacked on #955 (unknown-field tolerance): tolerance makes additive fields ignorable by updated parsers, and this PR adds the field plus the opt-in that keeps not-yet-updated parsers out of the blast radius entirely.

### Solution

* `RequestItem.disclose_claims: Option<bool>` — serde-optional and skipped when absent, so existing requests serialize byte-identically and old Authenticators never see the key. Includes a `with_disclose_claims` builder and `discloses_claims()` accessor, leaving `RequestItem::new`'s signature untouched.
* `ResponseItem.claims: Option<Vec<FieldElement>>` — populated via a new `with_claims` builder; `new_uniqueness`/`new_session` signatures unchanged.
* `generate_credential_proof` attaches `credential.claims` to the response item **only** when the request item opted in.

### Trust model note

The ZK circuit does not expose the claims hash as a public signal, so disclosed claims are not verifiable against the proof. They are authenticated by the transport-level app integrity signature (`integrity_bundle` version 2 binds nonce + per-item claims — golden vectors in `oxide/src/integrity/test.rs`). If proof-bound claims are wanted later, that's a circuit change (expose `claims_hash` publicly + RP-side Poseidon recompute) and intentionally out of scope here.

### Testing

* `cargo test -p world-id-primitives --lib` — 169 passed, including new round-trip tests: opt-in flag presence/absence, claims serialization/omission, and old-format requests parsing with disclosure off.
* `cargo check -p world-id-authenticator` — clean; clippy clean on both crates. (Authenticator/e2e suites need `forge` locally, so CI covers those.)

---

> [!NOTE]
> **Medium Risk**
> Introduces optional transmission of sensitive credential claims; RPs must not treat disclosed claims as proof-verified without the separate integrity binding, and misconfigured opt-in increases data exposure.
> 
> **Overview**
> Adds **opt-in disclosure** of raw credential claims on proof responses so RPs can request them without changing default wire behavior.
> 
> **Request side:** `RequestItem` gains optional `disclose_claims` (omitted unless `true`), plus `with_disclose_claims` and `discloses_claims()`. Existing requests stay byte-identical when the field is absent.
> 
> **Response side:** `ResponseItem` gains optional `claims` (`Vec<FieldElement>`) via `with_claims`; constructors default to no claims.
> 
> **Authenticator:** After building each response item, `generate_credential_proof` attaches `credential.claims` only when the matching request item opted in.
> 
> Disclosed claims are **not** bound by the ZK proof (no public claims hash); docs note they rely on transport-level app-integrity signing instead. Tests cover JSON round-trips and backward compatibility; call sites and fixtures set `disclose_claims: None`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 428414e7604f0f970f1f6fcc81a04fcbb01e5fdf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/authenticator/src/prove.rs
+++ b/crates/authenticator/src/prove.rs
@@ -458,6 +458,14 @@ impl Authenticator {
             )
         };
 
+        // Claims are disclosed strictly on RP opt-in; RPs that predate the
+        // field never receive it.
+        let response_item = if request_item.discloses_claims() {
+            response_item.with_claims(credential.claims.clone())
+        } else {
+            response_item
+        };
+
         Ok(response_item)
     }
 
--- a/crates/core/tests/generate_proof.rs
+++ b/crates/core/tests/generate_proof.rs
@@ -314,6 +314,7 @@ async fn e2e_authenticator_generate_proof() -> Result<()> {
             signal: Some(b"my_signal".to_vec()),
             genesis_issued_at_min: None,
             expires_at_min: None,
+            disclose_claims: None,
         }],
         constraints: None,
     };
--- a/crates/primitives/src/request/mod.rs
+++ b/crates/primitives/src/request/mod.rs
@@ -85,8 +85,13 @@ impl ProofType {
 }
 
 /// A proof request from a Relying Party (RP) for an Authenticator.
+///
+/// Unknown JSON fields are ignored during deserialization so that older
+/// Authenticators keep accepting requests from RPs speaking a newer minor
+/// revision of the protocol. The RP signature only covers the enumerated
+/// fields (see [`ProofRequest::digest_hash`]), so tolerated fields are
+/// never part of any signed or hashed message.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
 pub struct ProofRequest {
     /// Unique identifier for this request.
     pub id: String,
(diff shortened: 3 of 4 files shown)


paolodamico (changes requested):
claim disclosing is not supported for privacy considerations. claims are meant (must) be used within a zk circuit. happy to discuss more the motivation for this. further, the core protocol should not introduce outputs that are intended to be provable but can't be proved with the protocol (authenticator assertions do not exist on the protocol yet, https://github.com/worldcoin/world-id-protocol/pull/676)

SeanROlszewski:
Closing per review: claim values won't become protocol response outputs. The disclosure moved to the authenticator-vendor layer instead — WalletKit injects claims into its serialized response JSON on an opt-in generation path (worldcoin/walletkit#487), leaving world-id-primitives untouched. Thanks for the direction (and for merging #955).

PR #982, opened 2026-09-07
fix: reject recovery-revoked authenticators from registry state

alexblaniatfh:
Account recovery leaves stale packed mappings for the revoked authenticator. The indexer previously relied on an eventually consistent database counter, and the client classified revocation as an unregistered account, allowing `init_or_register` to create a replacement World ID.

Read the authoritative registry recovery counter, reject mismatches with HTTP 403 `authenticator_revoked`, and map that result to `PublicKeyNotFound` in both indexer and direct-RPC initialization. WalletKit already maps this error to `UnauthorizedAuthenticator`. A revoked authenticator never enters automatic registration. Existing clients that do not recognize the new error fail closed with an indexer/network error.

Validation: 20 authenticator unit tests pass; both packed-account integration tests pass against PostgreSQL 16 and Anvil, including stale and missing indexer rows; the real on-chain recovery test passes. Clippy passes for indexer/authenticator with all targets and indexer integration tests. Nightly formatting and diff checks pass. OHTTP regression coverage is added but its Docker-dependent test was not run locally. No contracts changed, no deployment performed. An opt-in local test database URL is documented; tests still create/drop only their own UUID databases.

Release dependency: publish the protocol crate and update WalletKit before relying on the specific revoked-device UI error. The indexer change independently closes the stale-database authorization gap.

AI disclosure: implemented and tested with OpenAI Codex.

### DIFF
--- a/crates/authenticator/src/authenticator.rs
+++ b/crates/authenticator/src/authenticator.rs
@@ -118,6 +118,8 @@ impl Authenticator {
     /// while a create-account or authenticator-management operation is still pending on-chain and
     /// the authenticator address has not been registered yet. Consumers that are coordinating such
     /// operations should poll the gateway request and retry initialization after finalization.
+    /// A recovery-revoked authenticator instead returns [`AuthenticatorError::PublicKeyNotFound`];
+    /// callers must request reauthorization rather than create a replacement World ID.
     ///
     /// Indexer DB catch-up is separate and does not block initialization, since packed account data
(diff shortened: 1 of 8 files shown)


Kemperino:
doesnt https://github.com/worldcoin/world-id-protocol/pull/968 and https://github.com/worldcoin/world-id-protocol/pull/968 take care of this already

PR #995, opened 2026-09-11
feat: benchmark WIP-106 circuit

kilianglas:
Makes the WIP-106 attestation provable end-to-end on the Noir/ProveKit backend and adds it to the mobile benchmark suite. Stacked on #884.

- Adds `authenticator-assertion-bench`, a reference consumer circuit over `authenticator_assertion::verify_attestation` that binds every caller obligation from `crates/proof/noir/README.md` as a public input. A new Noir CI job runs it and asserts the per-parameter visibility in the compiled ABI.
- All Rust machinery lives in `zk-mobile-bench` (unpublished), not in `world-id-proof`, so the crate's API cannot be mistaken for a new proof type: the bench crate's own build.rs compiles and embeds the prover, and a local module maps the witness. `world-id-proof` only makes the `sec_flags` / `authenticator_meta` claim packing pub.
- `zk-mobile-bench` gains `bench_authenticator_assertion_proof_generation`, reporting `prover_load` / `witness` / `prove` phases like the other Noir benchmark; the three caller workflows run it by default.

Test plan:
- [ ] `nargo execute --pedantic-solving` in `crates/proof/noir/authenticator-assertion-bench` solves the committed `Prover.toml`
- [ ] `cargo test -p zk-mobile-bench --release proves_and_verifies -- --include-ignored` — end-to-end prove + verify with a tampered-public-input negative control
- [ ] `cargo test -p zk-mobile-bench --release test_authenticator_assertion -- --include-ignored` runs the full benchmark path


kilianglas:
/mobench platform=both device_profile=all functions=zk_mobile_bench::bench_authenticator_assertion_proof_generation iterations=30 warmup=5

PR #1009, opened 2026-09-15
feat(relay): support AWS KMS signing, one key per network

karankurbur:
Add per-network AWS KMS signing to the relay while retaining legacy private-key support.
Set `AWS_KMS_SIGNING=true` and `{NETWORK}_AWS_KMS_KEY_ID`; false or unset allows `WALLET_PRIVATE_KEY`.
Reject conflicting or missing backends before provider setup; bound KMS calls and pin chain IDs.
Fund and authorize the KMS signers before deployment; ownership changes remain a separate cutover step.
Verified: 30 relay tests passed, one live fork test skipped; nightly clippy/fmt passed. Live KMS signing remains unverified.
Claude Code implemented KMS support; Codex fixed false-env rollback and added six isolated environment cases.


0xOsiris (commented):
Inherit this from [common::alloy::provider](https://github.com/worldcoin/world-id-protocol/blob/30afb896115329d5440ea78b05cfb8fc964c5cf2/services/common/src/alloy/provider.rs#L109) instead of re-writing

0xOsiris (changes requested):

karankurbur:
Done in e41b6d13 — `services/relay/src/signer.rs` is deleted and the relay now flattens `world_id_services_common::alloy::provider::SignerArgs`.

One thing worth surfacing, because it's why the copy existed: common pins the KMS signer to a chain id (`AwsSigner::new(.., Some(chain_id))` from the first RPC URL), and the relay signs on World Chain **and every satellite** with the same key. Alloy's `sign_transaction_with_chain_id!` returns `TransactionChainIdMismatch` when the signer's chain id differs from the transaction's, so a pinned wallet can sign World Chain and then fails on every satellite.

Rather than keep a fork, I added the missing case to common:

- `SignerArgs::chain_agnostic_wallet()` — same three inputs, built with `chain_id: None`, and no RPC endpoint needed since the chain id is exactly what's being skipped.
- both paths now share one `aws_kms_wallet_for_chain` builder, and the per-replica ordinal selection is factored into `per_replica_key`.
- that shared builder carries the bounded KMS timeouts and retries the relay copy had. **This changes gateway behaviour too**: the AWS SDK sets no operation timeout by default, so a hung `Sign` would previously block indefinitely. Now 3s connect / 5s per attempt / 15s total, 3 attempts with jittered backoff. Flagging explicitly in case you'd rather I scope it to the relay.

Also inherited, and documented in the README as not recommended here: `AWS_KMS_KEY_IDS` now parses for the relay. Per-replica keys would give each pod a different address, each needing funding on World Chain and every satellite.

`cargo clippy`, `cargo fmt` and the relay + common test suites are clean.

0xOsiris (dismissed):

0xOsiris (approved):

PR #1011, opened 2026-09-18
feat(gateway): durable wallet leases and restart-safe submission

Dzejkop:
Implements P2 of `services/gateway/WALLET_POOL_PLAN.md` (committed here). P1 (the app-lifecycle refactor from #949) is deliberately not included; it is independent plumbing and would have doubled the diff.

## The mechanism

A wallet may have at most one nonce-consuming transaction outstanding. Redis holds one record per wallet address:

- `gateway:wallet:{address}` — `signing` while a batch is being signed, then `in_flight` with the nonce, hash and **the signed bytes**, then `parked` if the fate cannot be decided. Every write is a compare-and-set Lua script; the attempt counter is part of the guard so concurrent resolvers cannot each advance it.
- `RequestRecord` gains an additive `wallet` field (`#[serde(default)]`). Records written by an older build decode with `None`, which is how the shrunk sweeper tells legacy submissions apart.
- `RequestStore::update_status_if` / `update_status_batch_if` apply a status only while the stored one is allowed. The batch form is all-or-nothing, so a transaction is never broadcast for a batch that was only partly recorded. This also closes a pre-existing race: the sweeper and the receipt tracker both wrote statuses blindly.

Submission order: acquire lease → sign → commit the signed transaction → transition the requests under the all-or-nothing guard → broadcast. If the lease was lost before the commit, or another owner resolved the requests, the signature is discarded rather than broadcast.

The resolver replaces the detached per-batch receipt task. It only concludes a transaction was replaced after re-reading its receipt on the same provider, because the provider fans out across RPC endpoints; an RPC failure is never evidence of absence; a receipt with no block hash waits rather than releasing below the confirmation floor. Undecidable transactions park their wallet: the requests are failed so clients get a bounded answer, but the lease is kept because releasing it could reuse a nonce whose transaction still exists.

The orphan sweeper keeps only requests no wallet record owns, and its terminal writes are guarded so it cannot overwrite a status another owner advanced. `tx_send_lock` is removed; the per-wallet lease already serialises sends.

## Compatibility

Additive only. No existing key is renamed or reshaped, and `update_status`'s whole-record `cjson` round-trip preserves the new field, so old and new builds can read and write the same records in both directions. The dual-shape window is bounded by existing TTLs rather than by a cutover — see §6.5.

## Verification

`cargo check`, `cargo clippy --all-targets --features integration-tests` and `cargo +nightly fmt --check` are clean. 30 gateway pure-logic tests and 29 `world-id-services-common` tests pass.

**Not verified:** the Redis-backed tests could not be executed here — this environment has no Docker socket for testcontainers, so 17 gateway tests (wallet store, request store guards, sweeper) are written but unrun. The resolver's probe matrix has no test at all; it needs a mock provider. Both are the first things to run in CI.

## Known gaps

- The lost-record reconciliation in §6.4 is specified but not implemented, and is documented as such; the primary control is a non-evictable Redis configuration.
- The throughput benchmark that §5.3 makes a go/no-go gate for a larger pool is not in this PR.

---

> [!NOTE]
> **High Risk**
> Changes the on-chain submission path (signing order, nonce safety, broadcast, and finalization) and Redis coordination; mistakes could double-broadcast, stall wallets, or mis-report request outcomes.
> 
> **Overview**
> Introduces **durable per-wallet transaction leasing** so the gateway never signs two batches from the same wallet at once, and can survive restarts without guessing about in-flight txs.
> 
> Batchers now only **build** unsigned `TransactionRequest`s (create vs ops via a shared `BatchType`); a new **`TransactionSubmitter`** acquires a Redis lease (`WalletStore`), signs, **write-ahead commits** the signed bytes, guarded-updates request status, then broadcasts. If no wallet is free within the acquire timeout, the batch is **re-queued** instead of failed. A background **resolver** owns receipt polling, re-broadcasts, confirmation counting, and **parking** wallets when fate is undecidable—replacing the old global `tx_send_lock` and per-batch receipt tasks on `RequestTracker`.
> 
> **Redis/request tracking** gains an optional `wallet` on `RequestRecord`, Lua **compare-and-set** status guards (`StatusGuard`), and configurable in-flight lock TTL. The **orphan sweeper** no longer hits the chain for receipts; it only fails **unowned** stale requests (legacy submissions without `wallet` still get a timeout path) and uses guarded writes so it cannot race the resolver.
> 
> **Configuration** adds flattened `WalletArgs` / `WalletConfig` (lease, resolution, rebroadcast, acquire timeout vs stale-queued bounds) plus wallet-related metrics. Startup wires **`http_wallet`** providers and spawns the resolver; tests and fixture spawns set `wallet: WalletArgs::default()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d622b3cd5817c9cc1c406e1831ef62b9887e70ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Dzejkop:
Superseded. This branch was split by commit order rather than by capability, so #1011 was pre-refinement code and #1012 carried the fixes to #1011 — which is why #1011's own tests failed on its branch and why the two shared comments kept conflicting.

Rebuilt as two independently-correct layers:

1. #1014 — guarded status writes
2. #1015 — wallet leases and resolution (based on #1014)

Closing in favour of those. The re-broadcast machinery was also removed in the rebuild: an absent transaction is terminal.

PR #1012, opened 2026-09-18
feat(gateway): share one wallet pool across replicas

Dzejkop:
Implements P3 of `services/gateway/WALLET_POOL_PLAN.md`. Stacked on #1011.

**This branch also carries three review-fix commits that touch #1011's code** (`829776cc`, `9004f8d5`, `d538de3f`). They are refinements of the mechanism, not of the pool: read them as review feedback on #1011 rather than as part of the pool change. The most important is the last one — a compare-and-set argument-order bug that made every successful wallet-record write fail, which the Redis-backed tests cover but could not be executed here.

## The pool

`AWS_KMS_WALLET_KEYS` and `WALLET_PRIVATE_KEYS` configure a comma-separated wallet pool that **every** replica is given, coordinated by the leases from #1011. `AWS_KMS_KEY_IDS` assigns one key per pod ordinal, so adding capacity meant editing a list whose length had to match the replica count — a comment in the infra repo asks a human to keep them in sync. A shared pool makes capacity one variable and lets pods and wallets scale independently.

`SignerArgs` gains the two variables inside its existing mutually-exclusive clap group, so a configuration that mixes a pool with a legacy variable is rejected at parse time and again in `validate()`. `ProviderArgs::http_wallets()` builds one wallet per pool entry through the existing single-signer path, so there is no second provider-construction path to trust; `http_wallet()` is unchanged.

`WALLET_DRAINING_ADDRESSES` excludes a wallet from new work while it keeps being resolved, so a wallet can be retired without abandoning a transaction it signed. Naming an unconfigured wallet, or draining all of them, is a startup error rather than a silent no-op.

Resolver passes are pinned to a single RPC endpoint, one read-only provider per configured URL, rotating per pass. The shared provider layer fans every call out across all URLs, so a receipt, a block and a nonce answered by different nodes cannot be reasoned about together.

## Rollout

Pool keys must be provisioned **disjoint** from the legacy per-ordinal keys, then switched in one template change; §8 has the argument. Adding wallets afterwards is a config edit.

## Verification

`cargo check`, `cargo clippy --all-targets --features integration-tests` and `cargo +nightly fmt --check` are clean. 29 `world-id-services-common` tests pass, including new coverage for `pool_signers` and `http_wallets`. The gateway's Redis-backed tests still cannot run here (no Docker).

**Before enabling a pool larger than today's:** run the §9 benchmark comparing `main`'s pipelined single wallet against N=1 and N=k leased at the same offered load. Serialising one transaction per wallet trades away pipelining, so a pool that is too small is slower than what it replaces. That benchmark is not in this PR because it needs a chain to run against.

---

> [!NOTE]
> **High Risk**
> Changes core on-chain submission, multi-replica wallet leasing, and transaction resolution—including dropping re-broadcast—so misconfiguration or resolver bugs could stall batches or mishandle in-flight transactions.
> 
> **Overview**
> Adds **shared wallet pools** so every replica can use the same signers via Redis leases, instead of tying one KMS key per pod ordinal. New env vars **`AWS_KMS_WALLET_KEYS`** / **`WALLET_PRIVATE_KEYS`** build multiple wallets through existing **`http_wallets()`** / **`pool_signers()`**; mixing pool and legacy signer config is rejected at startup. **`WALLET_DRAINING_ADDRESSES`** stops handing out selected wallets for new batches while the resolver still finishes their in-flight work.
> 
> Batch policy and the batcher now treat **`Batching`** requests as backlog (**`in_progress_count`**) so batches stuck waiting on a wallet are not treated as an empty queue or wrongly resynced from Redis.
> 
> The **transaction resolver** is reworked for multi-replica pools: one read-only RPC provider per URL per pass (rotating), a **supervised** resolver loop with pass timeouts, stricter guarded status/broadcast rules, cross-replica wallet acquire rechecks, and **removal of re-broadcast** (no stored raw tx / rebroadcast knobs). Undecided txs are probed with reorg-aware logic; truly absent txs fail requests and release the wallet instead of retrying broadcasts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73cb540416df2f1b8ec6421cb67f917ef4309031. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Dzejkop:
Superseded. This branch was split by commit order rather than by capability, so #1011 was pre-refinement code and #1012 carried the fixes to #1011 — which is why #1011's own tests failed on its branch and why the two shared comments kept conflicting.

Rebuilt as two independently-correct layers:

1. #1014 — guarded status writes
2. #1015 — wallet leases and resolution (based on #1014)

Closing in favour of those. The re-broadcast machinery was also removed in the rebuild: an absent transaction is terminal.

PR #1014, opened 2026-09-18
feat(gateway): guard request status writes

Dzejkop:
## What

Adds a compare-and-set status writer so a status write decided from a stale snapshot cannot overwrite a status another owner has already advanced.

- `StatusGuard` / `StatusWriteOutcome` and `RequestStore::update_status_if`, which applies a status only while the stored status is one of an allowed set. Terminal statuses still remove the request from the pending set and release its in-flight locks, matching `update_status`.
- `RequestTracker::set_status_if`.
- The orphan sweeper now fails a stale `Queued`/`Batching` request through the guarded writer, so a request another owner advanced between the snapshot and the write is left alone instead of being failed under it.

## Compatibility

Additive. The unguarded `update_status` writer is untouched, so existing call sites keep their current behaviour, and no Redis key, field, or shape changes.

## Tests

`guarded_status_write_only_applies_from_an_allowed_state` and `guarded_status_write_reports_a_missing_record`. Redis-backed, so they run in CI.

## Stack

1. **this PR** — guarded status writes
2. wallet leases and resolution — based on this branch

Part of making the wallet-submission work reviewable in independently-correct layers.

---

> [!NOTE]
> **Medium Risk**
> Changes concurrent request lifecycle semantics in Redis for the sweeper path; incorrect guard matching could leave stale requests unfailed or still allow races elsewhere that still use unguarded updates.
> 
> **Overview**
> Adds **compare-and-set (CAS) request status updates** in Redis so writers acting on a stale snapshot cannot overwrite a status another component already advanced.
> 
> `RequestStore` gains `StatusGuard` / `StatusWriteOutcome` and `update_status_if`, a Lua script that applies a new status only when the stored `status.state` is in an allowed set (`queued` / `batching`), preserves TTL, and on terminal writes still clears the pending set and in-flight locks like `update_status`. `RequestTracker::set_status_if` exposes this to the gateway.
> 
> The **orphan sweeper** now fails stale `Queued`/`Batching` requests via the guarded path instead of unconditional `set_status`, logging when the write is refused because the request moved on or when Redis errors.
> 
> Unguarded `update_status` is unchanged. New Redis-backed tests cover applied, guarded, and missing outcomes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88d5bc5f5dc47335be71ebd89c50cb811b000f73. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


services/gateway/src/orphan_sweeper.rs
```diff
@@ -79,15 +80,32 @@ pub async fn sweep_once(
                         "sweeper: failing stale {:?} request",
                         record.status,
                     );
-                    tracker
-                        .set_status(
```
josueberbesi2-cmd, line 83: 0xe8b4a8d5a617c56955d326e4e199257e683a81fa5466e5f955970c375f266484

PR #1019, opened 2026-09-21
fix: restore from cache

0xForerunner:
When restoring the tree from cache, we now only sync items from the database since the latest root in the cached tree. No need to resync db items from genesis.

---

> [!NOTE]
> **Medium Risk**
> Changes core Merkle tree startup and proof-serving correctness paths; mistakes could serve wrong roots or skip events, though bounded replay and post-replay root checks mitigate that.
> 
> **Overview**
> **Mmap cache restore** no longer replays registry events from genesis. After loading the cache, the indexer resolves the **latest** `RootRecorded` position for the cached root, replays only events up to the **latest committed root boundary** inside a **repeatable-read** DB snapshot, skips unchanged leaves, and **refuses to serve** unless the recomputed root matches that boundary. Failed restore still **deletes the cache** and exits.
> 
> Supporting DB APIs add **bounded** `get_after_until`, **`get_event_id_by_root`**, and **`get_latest_root_recorded`** (replacing boolean root existence). `TreeError` uses **`U256`** for root mismatch/missing; README documents the new startup and reorg-restart behavior. New **`test_cache_restore`** integration tests cover delta replay, snapshots, mismatches, and interrupted caches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77a31823e8ea23d0bc434c903e521c88d6cc3358. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


0xOsiris (approved):

services/indexer/src/tree/cached_tree.rs
```diff
@@ -344,7 +356,11 @@ async fn replay_events(
     );
 
     for (leaf_index, value) in &leaf_final_states {
-        set_arbitrary_leaf(&mut tree, *leaf_index as usize, *value);
+        let leaf_index = *leaf_index as usize;
+        if leaf_index < tree.num_leaves() && tree.get_leaf(leaf_index) == *value {
+            continue;
+        }
+        set_arbitrary_leaf(&mut tree, leaf_index, *value);
```
Dzejkop, line 363:
```suggestion
        // TODO: This function modifies the tree in-place. If a restart occurs mid apply the cache can end up with a root that doesn't match any of the RootRecorded events - which will cause a cache rebuild on the next start.
        set_arbitrary_leaf(&mut tree, leaf_index, *value);
```

